Issue #1, opened 2020-05-14
Route53 provider


danlsgiga:
Created this issue following the steps listed at [Implementing Providers](https://github.com/libdns/libdns/wiki/Implementing-providers)

danlsgiga:
@mholt, I also have the `caddy-dns` module ready... what's the best way to have it added to the caddy-dns org?

danlsgiga:
Went ahead and created https://github.com/danlsgiga/route53... can transfer ownership to the `caddy-dns` org whenever you want.

mholt:
@danlsgiga Great! This is fantastic. I got you all set up with the repositories and gave you maintain privileges on them.

Thank you!!

PR #2, opened 2020-05-14
Add the route53 libdns provider

danlsgiga:
This PR introduces the Route53 libdns provider

A few notes about the expected behaviours of the implementation:

- Route53 changes will return only after AWS validates the changes are `INSYNC`
- Since we are using the AWS Go SDK, all the authentication methods (Env vars, EC2 Instance Profile and Shared credential file) are supported out of the box
- Any application willing to use this library will need to initialize the AWS client session via `NewSession()`

WIP #1


route53/client.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	for _, rrset := range recordSets {
+		for _, rrsetRecord := range rrset.ResourceRecords {
+			record := libdns.Record{
+				Name:  *rrset.Name,
+				Value: *rrsetRecord.Value,
+				Type:  *rrset.Type,
+				TTL:   time.Duration(*rrset.TTL) * time.Second,
+			}
```
mholt, line 76: Is there any sort of "ID" associated with a record, from the Route53-side? i.e. does Route53 assign its own IDs to records? If so, it should be set in this struct too.

route53/client.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+	}
+
+	return "", fmt.Errorf("%s: No zones found for the domain %s", r53.ErrCodeHostedZoneNotFound, zoneName)
+}
+
+func (p *Provider) createRecord(ctx context.Context, zoneID string, record libdns.Record) (libdns.Record, error) {
+	// AWS Route53 TXT record value must be enclosed in quotation marks on create
+	if record.Type == "TXT" {
+		record.Value = strconv.Quote(record.Value)
```
mholt, line 119: I did not know about this function -- very cool!

... dang, I wish I knew about this earlier.

route53/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+[![godoc reference](https://img.shields.io/badge/godoc-reference-blue.svg)](https://pkg.go.dev/github.com/libdns/route53)
+
+This package implements the [libdns interfaces](https://github.com/libdns/libdns) for AWS [Route53](https://aws.amazon.com/route53/).
+
+## Authenticating
+
+This package supports all the credential configuration methods described in the [AWS Developer Guide](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html), such as `Environment Variables`, `EC2 Instance Profile` and the `AWS Credentials file` located in `.aws/credentials`
+
+The following IAM policy is a minimal working example to give Caddy permissions to manage DNS records:
```
mholt, line 12: One cool thing is that these libdns packages are totally unrelated to Caddy, so we don't need to make any references to Caddy; they can also be used by other code bases.

route53/client.go
danlsgiga: Yeah, I looked at that and they don't have any IDs for recordSets as per https://docs.aws.amazon.com/sdk-for-go/api/service/route53/#ResourceRecordSet

The way it works is you need to submit a [changeBatch](https://docs.aws.amazon.com/sdk-for-go/api/service/route53/#ChangeBatch) and in this struct you define all the the recordSet info... so for example, to DELETE a record, you submit a `changeBatch` with the `Name`, `Value`, `TTL` and `Type` of the recordSet and they do their sourcery under the hood.

route53/README.md
danlsgiga: Good point! Going to remove that reference

route53/client.go
mholt: Ah, interesting -- no problem then.

danlsgiga:
hmm... not sure... I'm getting access denied when I try to push there

mholt:
@danlsgiga It says the invite is still pending... need me to resend it?

danlsgiga:
nvm... I probably need more coffee... forgot to check my email! 😄 Thanks

PR #5, opened 2020-05-22
Initial commit TransIP provider

mdbraber:
Initial commit for TransIP. Straightforward implementation using the new v6 API. Building net/http function only would be possibly but likely look almost identical to the existing API which is more robust going forward IMHO. Also because TransIP offers different authentication methods (PrivateKey and Token) it would also be quite some extra work, while this implementation is rather lightweight.

I'm no expert Go coder, so appreciate all the feedback @mholt


transip/client.go
```diff
@@ -0,0 +1,97 @@
...
+	"context"
+	"time"
+
+	"github.com/transip/gotransip"
+	transipdomain "github.com/transip/gotransip/domain"
+	"github.com/libdns/libdns"
+)
+
+// NewSession initializes the AWS client
+func (p *Provider) NewSession() error {
```
mholt, line 13: This is not an AWS client, is it?

I just realized that the Route53 provider has this problem, in that it seems to require provisioning. Ideally, no provisioning would be required. Is this really necessary?

If so, it needs to be clearly documented.

Btw, if this is to be used by Caddy in the caddy-dns repositories, that type will need a `Provision()` method to call this for its Caddy module. I just raised a similar point over in route53: https://github.com/caddy-dns/route53/issues/2

transip/provider.go
```diff
@@ -0,0 +1,84 @@
...
+		newRecord.TTL = time.Duration(newRecord.TTL) * time.Second
+		appendedRecords = append(appendedRecords, newRecord)
+	}
+
+	return appendedRecords, nil
+}
+
+// DeleteRecords deletes the records from the zone. If a record does not have an ID,
+// it will be looked up. It returns the records that were deleted.
+func (p *Provider) DeleteRecords(ctx context.Context, zone string, records []libdns.Record) ([]libdns.Record, error) {
```
mholt, line 46: If TransIP doesn't assign IDs to records, then godocs about IDs don't make much sense IMO.

transip/client.go
mdbraber: I think this is just my lack of understaning how this should work. What's need is (obviously) a AccountName and a location of the PrivateKeyFile for the API to be connected. I couldn't directly figure this out how this should be done. Any pointers or some example code how this is suppoed to work in general?

transip/provider.go
mdbraber: No, this was a leftover from the Route53 implementation. Will remove

transip/client.go
mholt: Is calling `NewClient` expensive? If not, just make this unexported and have each of your exported methods call this when needed.

I'm only giving suggestions here, you're more familiar with TransIP of course :)

transip/client.go
```diff
@@ -0,0 +1,103 @@
...
+	"context"
+	"time"
+
+	"github.com/transip/gotransip"
+	transipdomain "github.com/transip/gotransip/domain"
+	"github.com/libdns/libdns"
+)
+
+func (p *Provider) setupRepository() error {
+	if p.repository == nil {
```
mholt, line 13: I think this is a good idea, but it just occurred to me that this isn't safe for concurrent use. (sync.Mutex would be helpful here.) The wiki doesn't mention this right now because I didn't think of it before, but I bet it'd be a good idea for each method to be safe for concurrent use.

That's my last bit of feedback. Otherwise, LGTM!

transip/client.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	return nil
+}
+
+func (p *Provider) getDNSEntries(ctx context.Context, domain string) ([]libdns.Record, error) {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+
+	p.setupRepository()
```
mholt, line 34: Deadlock? setupRepository will also try to obtain the same lock, which won't be released until this function returns.

transip/client.go
mholt: I updated the wiki and added the recommendations for thread safety, including the use of [sync.Once](https://golang.org/pkg/sync/#Once) if that makes things easier.

transip/client.go
mdbraber: Ah yes, that's stupid. I'm no mutex expert, would it be best to just have a mutex in each method and not in setupRepository?

mdbraber:
I'm happy to create a caddy-dns package too. I do have two concerns:

- I have no experience writing test, so some pointers would be appreciated (I couldn't find test with the other provider thus far)
- I tried testing for myself, but I have some questions regarding the expect inputs for libdns, see https://github.com/libdns/libdns/issues/7

mholt:
> I have no experience writing test, so some pointers would be appreciated (I couldn't find test with the other provider thus far)

Yeah, it's more like wishful thinking, since it primarily involves manipulating a remote resource, so mocks would be needed. For now, to just "ship it", at least try it out and make sure all the methods work for you.

mdbraber:
Shipped! https://github.com/libdns/transip

mholt:
Awesome!

mholt:
@mdbraber Just a low-priority reminder to fill out https://github.com/caddy-dns/transip when you get a chance -- should be mostly copy+paste from any other similar package. :+1:

mdbraber:
@mholt yes, it's definitely on my list and will get to it asap!

PR #14, opened 2020-09-24
OpenStack designate DNS provider

wolfedale:


openstack/pkg/designate/client.go
```diff
@@ -0,0 +1,172 @@
...
+func (p *Provider) deleteRecord(record libdns.Record, zone string) error {
+	err := recordsets.Delete(p.DNSClient, p.Request.ZoneID, p.Request.RecordID).ExtractErr()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func New(zoneName string) (*Provider, error) {
```
mholt, line 115: Is this optional? This isn't part of the API; specifically:

mholt (changes requested):
Thanks for working on this. This is pretty good! I've made some comments below, a few things will need to be changed before this is ready to add to the libraries.

Oh, and be sure to add godoc comments to all exported identifiers.

openstack/pkg/designate/client.go
wolfedale: Make sense. I will change it 👍

wolfedale:
I think I have fixed all those issues. Godoc comments are still missing.

wolfedale:
Does it look better now? :-)

openstack/pkg/designate/provider.go
```diff
@@ -0,0 +1,145 @@
...
+
+	var setRecords []libdns.Record
+
+	for _, record := range records {
+		recordID, err := p.getRecordID(record.Name, zone)
+		if err != nil {
+			return nil, err
+		}
+
+		p.setRecordID(recordID)
```
mholt, line 123: What is this for?

openstack/pkg/designate/client.go
```diff
@@ -0,0 +1,213 @@
...
+		return fmt.Errorf("cannot set environment variable: %v", err)
+	}
+	err = os.Setenv("OS_ENDPOINT_TYPE", p.AuthOpenStack.EndpointType)
+	if err != nil {
+		return fmt.Errorf("cannot set environment variable: %v", err)
+	}
+	return nil
+}
+
+func (p *Provider) auth(zoneName string) error {
```
mholt, line 152: This isn't thread-safe, client could authenticate multiple times.

openstack/pkg/designate/client.go
```diff
@@ -0,0 +1,213 @@
...
+func (p *Provider) deleteRecord(record libdns.Record, zone string) error {
+	err := recordsets.Delete(p.dnsClient, p.zoneID, p.recordID).ExtractErr()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *Provider) exportEnvVariables() error {
```
mholt, line 116: Does the underlying library _require_ env variables for configuration? If so, yuck; you'll need to make all uses of it mutually exclusive so that multiple configs can run on top of each other. It'll be very inefficient but there's no other way to do it correctly if that's the case. (They should fix that upstream.)

openstack/pkg/designate/provider.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/dns/v2/recordsets"
+	"github.com/libdns/libdns"
+)
+
+type Provider struct {
```
mholt, line 12: The exported fields will need JSON tags, for names like `auth_open_stack` -- and the AuthOpenStack type too, use `camel_case` for names.

openstack/pkg/designate/client.go
```diff
@@ -0,0 +1,213 @@
...
+		if recordName == rr.Name {
+			return rr.ID, nil
+		}
+	}
+
+	return "", nil
+}
+
+func (p *Provider) setRecordID(recordID string) {
+	p.recordID = recordID
```
mholt, line 56: I'm pretty sure this is a race condition

openstack/pkg/designate/client.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func (p *Provider) setDescription(desc string) error {
+	if desc == "" {
+		p.dnsDescription = "example description"
+	}
+	p.dnsDescription = desc
```
mholt, line 63: This is a race too. Why is all this information scoped to the Provider?

openstack/pkg/designate/client.go
```diff
@@ -0,0 +1,213 @@
...
+
+	dnsClient, err := openstack.NewDNSV2(provider, gophercloud.EndpointOpts{
+		Region: os.Getenv("OS_REGION_NAME"),
+	})
+	if err != nil {
+		return err
+	}
+	p.dnsClient = dnsClient
+
+	zoneID, err := p.setZoneID(zoneName)
```
mholt, line 176: Why is the zone set on the provider? The zone is scoped for each function call, not the lifetime of the provider struct.

mholt (changes requested):
Thanks for iterating on this, some more work is needed!

openstack/pkg/designate/client.go
wolfedale: Unfortunately from what I see they require env variables for the auth configuration :-(

openstack/pkg/designate/provider.go
wolfedale: added :-)

openstack/pkg/designate/provider.go
wolfedale: thanks for pointing it out, I have changed the code

openstack/pkg/designate/client.go
wolfedale: I added small check if client is already authenticated.

openstack/pkg/designate/client.go
wolfedale: I removed it completely.

openstack/pkg/designate/client.go
wolfedale: Thanks, good point. I moved it to the other method.

openstack/pkg/designate/client.go
mholt: Hmmm. I dunno if that's gonna fly. They don't expose _any_ way to manually configure with code?

wolfedale: I think I found a way :-)

wolfedale:
I think I've fixed most of my issues. Could you please check if it make sense now?

openstack/pkg/designate/provider.go
```diff
@@ -0,0 +1,161 @@
...
+// AuthOpenStack contains credentials for OpenStack Designate.
+type AuthOpenStack struct {
+	RegionName         string `yaml:"region_name"`
+	TenantID           string `yaml:"tenant_id"`
+	IdentityApiVersion string `yaml:"identity_api_version"`
+	Password           string `yaml:"password"`
+	AuthURL            string `yaml:"auth_url"`
+	Username           string `yaml:"username"`
+	TenantName         string `yaml:"tenant_name"`
+	EndpointType       string `yaml:"endpoint_type"`
```
mholt, line 28: Add `json` struct tags too if you could. (The majority of users, Caddy users, will need that.)

mholt (changes requested):
Cool, this is looking pretty good. You might need a mutex on the Provider struct to synchronize calls to `p.auth()`, for example. Otherwise, I don't see anything obviously wrong. Thanks!

mholt (changes requested):
The lock is precarious. I would move the locking into the `auth()` method and use defer to be sure it always gets unlocked. Right now, if it returns an error, it will never be unlocked, causing deadlocks.

wolfedale:
> The lock is precarious. I would move the locking into the `auth()` method and use defer to be sure it always gets unlocked. Right now, if it returns an error, it will never be unlocked, causing deadlocks.

Ok, now it make more sense. Thanks for telling me about this 👍

mholt (approved):
I would probably lock the whole function, rather than just one line -- otherwise, LGTM. I'll make the libdns/openstack and caddy-dns/openstack repos and give you permission to push when you're ready.

(Is there a better repo name you'd suggest?)

wolfedale:
> I would probably lock the whole function, rather than just one line 
Done :-)
 
> (Is there a better repo name you'd suggest?)
I'm wondering it "designate" is not a bit better name. [https://wiki.openstack.org/wiki/Designate]

mholt:
How about `openstack-designate` then?

wolfedale:
> How about `openstack-designate` then?

Sounds good to me :-)

Issue #15, opened 2020-10-06
HetznerDNS provider


matthiasng:
PR coming soon.

mholt:
Awesome, looking forward to it!

afoo:
@matthiasng I'm also interested in having this, let me know if I can help

matthiasng:
@afoo ty. Im nearby done but feel free to test/review the PR. At the moment i only have to impl and test the pagination for get records/zones

PR #16, opened 2020-10-10
Hetzner provider

matthiasng:
PR for #15 15

### DIFF
--- a/hetzner/provider.go
+++ b/hetzner/provider.go
@@ -7,7 +7,7 @@ import (
 	"github.com/libdns/libdns"
 )
 
-// Provider implements the libdns interfaces for DigitalOcean
+// Provider implements the libdns interfaces for Hetzner
 type Provider struct {
 	//Client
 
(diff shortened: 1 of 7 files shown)


hetzner/provider.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"context"
+	"strings"
+
+	"github.com/libdns/libdns"
+)
+
+// Provider implements the libdns interfaces for Hetzner
+type Provider struct {
+	//Client
```
mholt, line 12: Possible cleanup needed on aisle 12 :)

mholt (approved):
This actually looks pretty good from what I can see!

I'm creating the hetzner repo and adding you to it so you can push to it and maintain this. Thanks again!

Unrelated, but if you'd be interested in helping all CertMagic + Caddy users be able to use this, a simple wrapper over this provider could be created in the caddy-dns organization: https://github.com/caddy-dns -- would you be able to do that too? If so I will create that repo and add you to it as well. :) They're pretty simple; the "hardest" thing is just Caddyfile support which you can mostly copy-paste from any other repo. Anyway, you're totally not obligated to do this. Just that most users here tend to also be Caddy users right now so I thought I'd ask.

matthiasng:
I'm always glad to help. Just create the caddy repo and send me an invite. I will do the rest ;)

mholt:
Thanks! Done

Issue #24, opened 2021-01-25
Azure DNS Provider


kurokobo:
Add new provider for [Azure DNS](https://azure.microsoft.com/en-us/services/dns/).

kurokobo:
@mholt 
Thank you for your support.
I've pushed my codes in [libdns/azure](https://github.com/libdns/azure) and [caddy-dns/azure](https://github.com/caddy-dns/azure) as `v0.1.0`, and I've tested those new repos work well with Caddy running in my container :)

Is there anything I need to work on to add Azure provider to the [download page](https://caddyserver.com/download) and [module list](https://caddyserver.com/docs/modules/)?
If there is nothing in particular for me, this issue is fine to close.

mholt:
Cool, looks great!

To add it to the download page, just log in to your Caddy account (sign up if you don't have one already) and then click Register Package and fill out the couple of fields.

kurokobo:
Thank you, I have completed the registration of the package. It's all done 👍

PR #25, opened 2021-01-25
Add Azure DNS Provider

kurokobo:
#24

These implementation (also available at [my repo](https://github.com/kurokobo/libdns-azure)) and related [caddy-dns module](https://github.com/kurokobo/caddy-dns-azure) is already working with Caddy in my private environment for my personal use.
It would be great if you could review this PR.


kurokobo:
I forgot to change the name and year in the LICENSE :(
I just fixed it so it's ready for review now, thank you.

mholt:
Thanks for implementing a provider!

I see a few things that will need to be resolved or clarified. First, see [the requirements here](https://github.com/libdns/libdns/wiki/Implementing-providers#requirements):

> Configuration must be primarily through struct fields. Environment variables are allowed, but should only be used as defaults.

It looks like you accept env variables but do not require them, is that right? If so, that's OK; as long as it be configured without env vars.

I see there's a `NewClient()` method, unfortunately this won't fly (if I am understanding it right). It has to be able to be used without a separate function call first. Other providers often do this by calling a function (synced by a mutex) at the beginning of each exported method to ensure the client is set up before it tries to do anything.

At a glance, tests and other things look good though!

kurokobo:
Thanks for taking your time and the quick review!

> It looks like you accept env variables but do not require them, is that right? If so, that's OK; as long as it be configured without env vars.

No, all configurations need to be explicit in structure fields. My implementation uses `azure-go-sdk` and it has the ability to authenticate using environment variables, but my implementation does not use that feature.
I think my README and example might be the cause of the confusion. I'll fix the description.

> I see there's a NewClient() method, unfortunately this won't fly (if I am understanding it right).

Ah, you are right.
I was referring to the route53 implementation (that has `NewSession()`), but now I understand your concern. I will fix it.

Please give me some time to fix it. Thank you.

kurokobo:
I've fixed that. Could you please check this again?

mholt (approved):
Looks much better! Thanks for clarifying/fixing.

I'll make a repo for this at `libdns/azure` -- sound good?

Also, would you be willing to contribute your Caddy plugin to the `caddy-dns/azure` repo which I will make soon?

kurokobo:
Thanks for your review! I'm glad it's approved :)

I have received the two invitations. I will push the code to both libdns and caddy-dns in probably today, or in a few days at the latest.

Issue #26, opened 2021-02-16
dynv6 provider


sstamoulis:
Creating libdns provider for [dynv6 REST API](https://dynv6.github.io/api-spec/)

PR #27, opened 2021-02-16
dynv6 provider

sstamoulis:
This is PR for #26. In order to run tests you will need to provide a token for authorization using the -token flag.


sstamoulis:
This library now works fine using caddy on a dev server

provider.go
```diff
@@ -0,0 +1,124 @@
...
+	"fmt"
+
+	"github.com/libdns/libdns"
+)
+
+// Provider for dynv6 HTTP REST API
+type Provider struct {
+	// Token is required for authorization.
+	// You can generate one at: https://dynv6.com/keys
+	Token string
```
mholt, line 14: Add JSON tag:
```suggestion
	Token string `json:"token,omitempty"
```
Allowing for easy JSON configuration :+1:

mholt (approved):
Cool, looks pretty good -- just a nit about JSON struct tag for configuration. It's great that it has tests!

I'll make a repo for it, and a caddy-dns repo as well if you would be able to push the DNS module there. :)

Make sure to apply a compatible OSS license when you push!

Thanks for making this!

sstamoulis:
Hey ,thanks for the suggestions. Already updated the code and license.
I do not have permission to push to libdns/dynv6 though.
Could you maybe take a look?

mholt:
Make sure to accept the invitation. It says "Awaiting sstamoulis’s response"

rishubn:
@mholt Why was this never pushed? libdns/dynv6 is empty

mholt:
@rishubn I dunno, ask @sstamoulis  -- the invite is still pending. :man_shrugging:

rishubn:
Since the code is licensed under MIT, would you object if I submitted a
fork (with correct attributions)? I'd agree to maintain it.

Of course if they decide to return, I will defer to their wishes.

mholt:
Sure, we welcome forks of unmaintained plugins. I'll add you to the repo (and also caddy-dns/dynv6).

PR #29, opened 2021-03-17
Add Metaname provider

mwh:
This adds a provider for [Metaname](https://metaname.net/) domain name services.


mholt (approved):
Awesome, looks good. I'll create the repo and invite you to it.

Would you be willing to create a caddy-dns module as well, so that Caddy users can benefit from this provider? (They're pretty easy: https://github.com/caddy-dns/template)

mwh:
Yes, that's the plan.

PR #33, opened 2021-04-18
Namedotcom

justenb:
A tested name.com libdns implementation. Please review and thanks for all your work.


provider.go
```diff
@@ -0,0 +1,81 @@
...
+package namedotcom
+
+import (
+	"context"
+	"github.com/libdns/libdns"
+)
+
+// Provider implements the libdns interface for namedotcom
+type Provider struct {
+	Client
```
mholt, line 13: This probably doesn't need to be embedded/exported does it? Nothing within the Client type is exported, as far as I can tell.

mholt (approved):
Great, LGTM -- thanks for your contribution! I'll make a libdns/namedotcom repo as well as caddy-dns/namedotcom repo (assuming you'd be interested in using this with Caddy? making these is really easy) for you to push to.

provider.go
justenb: Thanks for reviewing!

It doesn't need to be exported. I have corrected this in the latest commit. (https://github.com/justenb/libdns/commit/2c725af50ba1f98b2d4ee322e1fc0eeb03ac306d)

I will follow up with pushing the latest to the libdns/namedotcom repo along with adding the caddy module.

mholt: Awesome, thanks again for contributing this!

PR #36, opened 2021-05-17
Leaseweb

### LINKED ISSUE #35: Leaseweb Provider
Implementing [Leaseweb](https://leaseweb.com/) provider.

justinvdk:
Implementation of the Leaseweb provider.
First go code so critical feedback welcome.

Closes #35

### DIFF
--- a/models.go
+++ b/models.go
@@ -1,0 +1,14 @@
+package leaseweb
+
+// Structs for easy json marshalling.
+// Only declare fields that are used.
+type LeasewebRecordSet struct {
+	Name    string   `json:"name"`
+	Type    string   `json:"type"`
+	Content []string `json:"content"`
+	TTL     int      `json:"ttl"`
+}
+
+type LeasewebRecordSets struct {
+	ResourceRecordSets []LeasewebRecordSet `json:"resourceRecordSets"`
+}
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) <TODO: YEAR AND NAME>
+Copyright (c) 2021 Justin van der Krieken
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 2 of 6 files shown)


models.go
```diff
@@ -0,0 +1,14 @@
...
+type LeasewebRecordSet struct {
+	Name    string   `json:"name"`
+	Type    string   `json:"type"`
+	Content []string `json:"content"`
+	TTL     int      `json:"ttl"`
+}
+
+type LeasewebRecordSets struct {
+	ResourceRecordSets []LeasewebRecordSet `json:"resourceRecordSets"`
+}
```
mholt, line 14: FYI, if these aren't used outside of this package, you can un-export them (start with lowercase, like `leasewebRecordSet`).

mholt (approved):
Nice job! I don't see any glaring issues. I will create the new repository and invite you to it.

If you're using this with Caddy, would you also be willing to add a caddy-dns module? https://github.com/caddy-dns

models.go
justinvdk: Thanks! I'll fix this in the repo.

mholt: Cool; same goes for any identifier that isn't needed publicly. General rule is, if it's only used within the package, keep the type name lowercase to avoid increasing your exported API surface. :+1:

justinvdk: Makes sense! New to go so still need to get used to the syntax. I'll also look into the Caddy module, I was planning to check Caddy out sometime anyway 👍🏻

mholt: Neat. You might be the first libdns consumer that isn't using Caddy already -- what do you use libdns for, being new to Go and such?

justinvdk: Oh, now that you mention it. I was confused with Traefik. Yes I started this implementation for Caddy. Oops.

I'm looking into building caddy with my provider with xcaddy as we speak 😄

Issue #41, opened 2021-06-02
Google Cloud DNS Provider


brockwood:
I'm planning on implementing the [Google Cloud DNS](https://cloud.google.com/dns) provider.

PR #42, opened 2021-06-03
Add the Google Cloud DNS provider

brockwood:
This PR introduces the Google Cloud DNS libdns provider

A few notes about the expected behaviours of the implementation:

* This is based on the Google Cloud Go library, all the authentication methods (environment variables, service account JSON, VM metadata) are supported out of the box
* An application using this library will need to initialize the Google Cloud client session via `NewService`
* The tests are built around the `httpreplay` library included in the `cloud.google.com/go` library and the included JSON files are the data for each test containing all the calls made for that test
* A single Google Cloud recordset is broken into several libdns.Records due to a single record containing multiple values (see example recordset JSON)

Here's a JSON representation of a recordset returned from Google:
```json
{
  "kind": "dns#resourceRecordSet",
  "name": "example.com.",
  "rrdatas": [
      "10 mail.example.com.",
      "20 mail2.example.com."
  ],
  "ttl": 86400,
  "type": "MX"
}
```

Provider issue #41


googledns/client.go
```diff
@@ -0,0 +1,137 @@
...
+	"google.golang.org/api/option"
+)
+
+const (
+	// zoneMapTTL timeout for the Google Cloud DNS zone map
+	zoneMapTTL = time.Minute * 5
+)
+
+// NewService initializes the Google client for the provider using the specified JSON file for credentials if set.
+func (p *Provider) NewService(ctx context.Context) error {
```
mholt, line 20: Unfortunately, this will be a deal-breaker for inclusion in the libdns repositories; one of the [requirements](https://github.com/libdns/libdns/wiki/Implementing-providers) is that no additional provisioning steps are necessary:

I think a few other implementations just do it implicitly on first use.

mholt (changes requested):
Awesome, the implementation looks good, _except_ for the extra provisioning step. (If it's optional, that's probably allowable, but it can't be required.) Rework that please and then we can make a repo for this!

Would you also be willing to add the caddy-dns wrapper if I make a repo for it?

googledns/client.go
brockwood: Understood 👍 . I noticed other packages were using `sync.Mutex` so It shouldn't be difficult matching that pattern. Let me get that change in and give it a quick test.

mholt: Yep, sure; or `sync.Once()` if you prefer.

brockwood:
> Would you also be willing to add the caddy-dns wrapper if I make a repo for it?

You bet. I've already been testing a local Caddy build with a caddy-dns wrapper for this provider so I'd be glad to contribute that as well.

mholt:
Cool. Do you think the repositories should be named `googlecloud` or `google-cloud`?

brockwood:
> Cool. Do you think the repositories should be named `googlecloud` or `google-cloud`?
> 
> Edit: I just realized, the package name here is `googledns` -- is "Google DNS" the canonical name for this service? Or should it be `googleclouddns` or something? (I'm not saying it needs to change, just a lastt minute thought.)

🤔 Hmmm... I actually like being more explicit with `googleclouddns`. Let me get that change in.

googledns/client.go
brockwood: The service provisions on first use without having to call `newService` so we should be good to go there.

mholt (approved):
Excellent, let's get this pushed to the libdns/googleclouddns repo! Thanks for your contribution! I think a lot of people will find it useful.

Issue #60, opened 2022-02-20
Porkbun provider


Niallfitzy1:
Port of [Porkbun](https://go-acme.github.io/lego/dns/porkbun/) 

I'll make an attempt at it myself

erri120:
What is the status of this? #61 has been closed, but https://github.com/libdns/porkbun and https://github.com/caddy-dns/porkbun are empty.

Niallfitzy1:
Thanks for the reminder, this completely slipped my mind.

Pushed the repos this evening: 
[libdns/porkbun v0.1.0](https://github.com/libdns/porkbun/releases/tag/v0.1.0)
[caddy-dns/porkbun v0.1.0](https://github.com/caddy-dns/porkbun/releases/tag/v0.1.0)

mholt:
Awesome, thanks!

PR #61, opened 2022-02-23
Implement Porkbun provider

Niallfitzy1:
I have a provider for Porkbun working, the readme details the test process which doubles as a usage example.

Looks after #60 

I've also got a DNS Wrapper module [here](https://github.com/Niallfitzy1/caddy-dns-porkbun) that I'm using for now


mholt:
Thanks for the PR -- just catching up from being sick, will get to this soon.

_test/test.go
```diff
@@ -0,0 +1,98 @@
...
+		return
+	}
+
+	provider := porkbun.Provider{
+		APIKey:       apikey,
+		APISecretKey: secretapikey,
+	}
+
+	//Check Authorization
+	_, err := provider.CheckCredentials(context.TODO())
```
mholt, line 32: If this is required, it won't meet our criteria of no extra setup or provisioning steps allowed (since the providers need to be interchangeable).

provider.go
```diff
@@ -0,0 +1,371 @@
+// Package libdnstemplate implements a DNS record management client compatible
+// with the libdns interfaces for Porkbun.
+package libdns_porkbun
```
mholt, line 3: In the final publication, call the package just `porkbun` (or whatever you want, but `libdns_` is not a necessary / conventional thing).

mholt (approved):
Looks good overall, just had a couple nits. I'll assume you'll address those and create the libdns repo and the caddy-dns repo, too, and invite you to push your code there. Thanks!

go.mod
```diff
@@ -1,3 +1,7 @@
-module github.com/libdns/libdns
+module github.comn/Niallfitzy1/libdns-porkbun
```
mholt, line 1: Ensure this is `github.com/libdns/porkbun` before you push and tag there

_test/test.go
Niallfitzy1: Not required, it just checks credentials are valid which was handy when setting things up initially

Issue #63, opened 2022-03-23
Joohoi's ACME-DNS provider


vbsd:
ACME-DNS is a simplified DNS server that only allows setting TXT values. Only up to two records exist at a time, the older ones get deleted automatically. It is meant to be used for solving ACME DNS challenges. More information:

* https://github.com/joohoi/acme-dns
* https://www.eff.org/deeplinks/2018/02/technical-deep-dive-securing-automation-acme-dns-challenge-validation

Any implementation of this provider will be extremely limited. However, Caddy and `certmagic` only uses `RecordAppender` and `RecordDeleter` interfaces for obtaining certificates (and it's okay if `RecordDeleted` returns an error), so it would still be useful.

PR #64, opened 2022-03-23
Implement Joohoi's ACME-DNS provider

vbsd:
Related issue: [#63](https://github.com/libdns/libdns/issues/63)

ACME-DNS is quite unusual in its goals - it is very limited, and so is this `libdns` provider. More information on that is in the issue description and in `README.md` of this PR. My goal is to use this for a Caddy DNS plugin. Let me know if you think it would be better to _not_ put this code into a separate `libdns` repo and just include it in Caddy DNS plugin module (which I would like to host under `caddy-dns` organization eventually).

Any comments and suggestions for improvement is appreciated, I'm fairly new to Go.


vbsd:
I made some significant changes in the last commit. Instead of using just one ACME-DNS account, the configuration now matches one-to-one with configuration used by [acme-dns-client](https://github.com/acme-dns/acme-dns-client), which means that there's a `domain -> domain config` mapping. `AppendRecords()` uses zone and record name information to select which config/account to use.

provider.go
```diff
@@ -0,0 +1,95 @@
+package joohoi_acme_dns
```
mholt, line 1: Might be best to stick with Go convention, that is the name of the folder, and no underscores:

```suggestion
package acmedns
```

provider.go
```diff
@@ -0,0 +1,95 @@
...
+
+type DomainConfig struct {
+	Username   string `json:"username,omitempty"`
+	Password   string `json:"password,omitempty"`
+	Subdomain  string `json:"subdomain,omitempty"`
+	FullDomain string `json:"fulldomain,omitempty"`
+	ServerURL  string `json:"server_url,omitempty"`
+}
+
+type Domain = string
```
mholt, line 24:
```suggestion
type Domain string
```

(Could probably just use a `string` too, since there's no behavior associated with this type.)

provider.go
```diff
@@ -0,0 +1,95 @@
...
+		appendedRecords = append(appendedRecords, libdns.Record{Type: "TXT", Name: record.Name, Value: record.Value})
+	}
+	return appendedRecords, nil
+}
+
+// Implements libdns.RecordDeleter.
+//
+// Always returns and error since ACME-DNS API does not support record deletion.
+func (p *Provider) DeleteRecords(ctx context.Context, zone string, recs []libdns.Record) ([]libdns.Record, error) {
+	return nil, fmt.Errorf("joohoi_acme_dns provider does not support record deletion")
```
mholt, line 88: What was the result when you tested this? I feel like this would produce errors in the logs and I'm not sure that's desirable/necessary in the case of ACME challenges (which is all this can be used for anyway), if it cleans up the TXT records for us.

mholt (changes requested):
Thanks for implementing this! I know we've had a few requests for this so far. Just a few nits/questions before I recommend it for its own repository. Thank you!

provider.go
vbsd: I just wanted to make clear that the mapping is from domain to `DomainConfig` But I guess that's clear enough from the docs and context, I will just go with `string`.

provider.go
mholt: Good question! I'm not sure, haha. I don't know of any implementation that actually uses the return value for DeleteRecords. Hmm, maybe return `nil, nil`? I dunno. Perhaps be prepared to change that if someone complains. :man_shrugging:  I think this will only have a good answer after some experience in the wild.

But I think an error will almost certainly lead to "why is it not working" questions, when it is actually working just fine.

vbsd: Makes sense, changed it to `return nil, nil`. :+1:

provider.go
vbsd: Changed: https://github.com/libdns/libdns/pull/64/commits/fbf6d6cdc86fd94ceec785b4dcb741319d4511c0

mholt (approved):
LGTM, thank you!

I'll create a libdns/acmedns and caddy-dns/acmedns repo and add you to them, so you can push to those. Make sure your go.mod file matches of course :) Thanks!

vbsd:
Thank you for quick response! :)

vbsd:
@mholt 
What is the process for accepting [caddy-dns/acmedns](https://github.com/caddy-dns/acmedns)? I've pushed the implementation, could you also have a look at it to check if it seems reasonable?

mholt:
@vbsd Looks good to me! (There's no process for that repo.)

vbsd:
Great, thanks you!

PR #77, opened 2022-06-21
RFC2136 Provider

KoHcoJlb:
You can see in `conversion_test.go` how TXTs are working.


KoHcoJlb:
My concern was that quoting could break some existing code, if it's good for you it's good for me.

Thanks for Caddy:) Still exploring it but it's already looking to me like very powerful multitool http server.

mholt:
Cool. And yeah, like I said, let's just give it a try like this, and if it causes problems, we can fix/change it!

Issue #85, opened 2022-08-30
Civo Provider


dirien:
Hi,

I created the libdns provider for Civo. -> https://github.com/dirien/civo

Will create an PR here to add the provider repo.

mholt:
Thank you -- reviewed!

PR #86, opened 2022-08-30
feat: add Civo provider

dirien:
This PR adds Civo as provider.

Closes issue #85


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020 Matthew Holt
+Copyright (c) <TODO: YEAR AND NAME>
```
mholt, line 3: (Update this)

mholt (approved):
At a quick glance, looks good. Similar to the scaleway one. Thanks for contributing! Again make sure record names are zone-relative, not absolute. (But convert as needed for civo's API)

LICENSE
dirien: Will do! Thanks for pointing out!

Issue #88, opened 2022-10-06
Claiming nja.la


katzeprior:
Claiming nja.la, going to add it here for caddy support for a friend who needs it :)

mholt:
@katzeprior https://github.com/libdns/libdns/compare -- see GitHub docs: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request

katzeprior:
@mholt I know how to do normal pr requests, is my repo correct?
![image](https://user-images.githubusercontent.com/87608942/194427044-66fc1c61-8918-4a1c-94ea-c92f480a4bd1.png)
Can't select my katzeprior/njalla repo here
![image](https://user-images.githubusercontent.com/87608942/194427145-ea537735-fa99-48dd-8a0b-7704d7a6bd5c.png)

mholt:
You can make a PR to the template repo I guess. The point is just that I need a page where I can review the code.

PR #89, opened 2022-10-06
add: working functionality

katzeprior:
Adding njalla support as stated in issue: #88 for caddy support :)

I see what I did wrong with the initial forking and not being able to create a pr.


njalla.go
```diff
@@ -0,0 +1,185 @@
...
+		return libdns.Record{}, err
+	}
+
+	result := struct {
+		Result NjallaRecord `json:"result"`
+	}{}
+	if err := json.Unmarshal(data, &result); err != nil {
+		return libdns.Record{}, err
+	}
+	log.Println(result)
```
mholt, line 149: Remove these log/debug lines

njalla.go
```diff
@@ -0,0 +1,185 @@
...
```
mholt, line 109: Remove this

mholt (approved):
Thanks for contributing this; just a couple stray logs to be removed.

I'll create a repo for you in libdns and in caddy-dns, since I understand this is also for Caddy use. You can push whenever you accept the invite to those repos. Thanks again!

PR #91, opened 2022-11-28
Total Uptime provider for libdns - resolves #90

### LINKED ISSUE #90: New provider: Total Uptime
Hello, I'd like to build an integration with one of our providers Total Uptime.
Thanks!

kmei3560:
Hello, this PR is an implementation of libdns for Total Uptime DNS provider.

Please let me know if I've not submitted this correctly. Also, feel free to give any feedback on the code itself, or let me know if I've done something wrong.

Thank you


totaluptime/README.md
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+
+	"github.com/libdns/totaluptime"
+)
+
+func main() {
+	var provider totaluptime.Provider
+
+	provider.Username = USERNAME      // provider API username
+	provider.Password = PASSWORD      // provider API password
```
mholt, line 45:
```suggestion
	provider := totaluptime.Provider{
		provider.Username = "USERNAME",      // provider API username
		provider.Password = "PASSWORD",      // provider API password
	}
```

I think this is more typical convention

totaluptime/examples/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/kmei3560/libdns/totaluptime/examples
```
mholt, line 1: Don't forget to change this when you push to the libdns repo

mholt (commented):
Thanks for the submission!

Overall pretty good. I would recommend getting rid of the Print lines that aren't in the example file; generally you should return the first error right away so that they can be fixed.

Let me know when you're ready and I'll make the repos for you to push to. Thanks!

totaluptime/README.md
kmei3560: Got it. Thanks

kmei3560:
Can you please take another look when you have a chance?

Also, I don't understand what you mean by this statement:

> I would recommend getting rid of the Print lines that aren't in the example file; generally you should return the first error right away so that they can be fixed.

Thanks

kmei3560:
Please review latest change and see if this is better (I agree it's more proper with `log.Fatalln`)
If everything looks good, you can go ahead and setup the new repo. Thanks!

totaluptime/provider.go
```diff
@@ -0,0 +1,337 @@
...
+		return nil, fmt.Errorf("record append cannot proceed: unknown zone: %s", zone)
+	}
+
+	// append each record individually
+	for _, rec := range records {
+		// identify record type
+		typeName, err := convertRecordTypeToProvider(rec.Type)
+		if err != nil {
+			log.Printf("Unable to identify record type; skipping: %s: %s\n", rec.Type, rec.Name)
+			continue // skip to next record
```
mholt, line 158:
```suggestion
			return nil, err
```

There's a lot of lines like this... ideally, don't emit logs -- just do, or do not. (There is no try.)

I'm not familiar with TU, so it's up to you decide: if this case is unexpected, then return the error. If it's normal and harmless (doesn't affect the final result; i.e. this is a debug-level log), then remove the log and just continue.

mholt (commented):
(Sorry, been catching up after the holiday, and now taking care of sick kids.)

Hopefully this clarifies things! Basically, try to figure out of each of these cases is normal and OK or not. If not, return an error. If so, don't print a log.

kmei3560:
Gotcha. It makes a lot of sense to me when you say "don't emit logs"... yes I tend to be overly verbose with logging.
I tried to eliminate the scenarios you mentioned (DO or DONT, but don't emit logs unnecessarily)

There may be a few more cases remaining, so just let me know if I haven't taken this far enough. Typically I like to log a condition that caused something unexpected. So for example, if an incorrect domain was requested to be deleted - I would want to log that the domain could not be found (to help user understand).

Thank you very much for your Go wisdom!  Very much appreciated! 
Also, sorry to hear about sick kids. Hope everyone feels better soon!

mholt (approved):
Great, looks better now :) Thanks for the contribution! I'll create the repos and add you to them. (libdns and caddy-dns as well)

kmei3560:
Sounds good - and thank you!  For the record, I do not use `caddy` as-of-yet. I currently only use `certmagic` library and `libdns` providers.

mholt:
Ah, apologies for presuming. (You're probably the first who contributed a provider but isn't using Caddy.) I'll leave the repo and invite there, but don't feel obligated!

PR #94, opened 2023-01-30
Introduction of a Linode provider for libdns

bcrochet:
Uses the v4 Linode API via linodego library.


linode/client.go
```diff
@@ -0,0 +1,136 @@
...
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/libdns/libdns"
+	"github.com/linode/linodego"
+)
+
+func (p *Provider) init(ctx context.Context) {
+	p.client = linodego.NewClient(http.DefaultClient)
```
mholt, line 15: This probably only needs to happen once (and should be synced to be thread-safe; use a mutex, probably)

mholt (approved):
LGTM, just one little nit about synchronization.

I'll create a repo and let you push to it. Thank you for the contribution, and for maintaining it! I'll also create a caddy-dns repo for this provider so you can optionally have it plug in to Caddy.

linode/client.go
bcrochet: I'll update it. I want to add some unit tests too. Thanks for the feedback.

PR #100, opened 2023-05-03
New Provider: Direct Admin

rnavarro:
This PR adds support for Direct Admin


client.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	if len(respData.Error) > 0 {
+		return libdns.Record{}, fmt.Errorf("%v: %v", respData.Error, respData.Result)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		bodyBytes, err := io.ReadAll(resp.Body)
+		if err != nil {
+			log.Fatal(err)
```
mholt, line 292: Never crash -- return errors instead.

mholt (changes requested):
Thanks! Overall LGTM, but very important to not crash programs. Instead, return errors.

I'll go ahead and create a libdns/directadmin repo -- would you be able to make it a Caddy module as well? If so I'll create the caddy-dns/directadmin repo.

Don't forget to update your go.mod to the correct package path.

rnavarro:
> would you be able to make it a Caddy module as well? If so I'll create the caddy-dns/directadmin repo.

The original intent was to use DirectAdmin in Caddy, which is how I fell down the rabbit hole, so I think the answer is yes :)

mholt:
Fantastic. Thanks for contributing this! I've added you to both repos, so feel free to push when ready.

Issue #104, opened 2023-06-26
EasyDNS Provider


sdowns:
I'd like to create a provider for [EasyDNS](https://easydns.com).

mholt:
@sdowns Awesome, would absolutely love to add that!

PR #105, opened 2023-06-30
Add EasyDNS Provider

sdowns:
Adds EasyDNS Provider per issue #104.


mholt (approved):
Just did a quick scan and it LGTM. Thanks for contributing!

I'll make you a repo for you to push this to, and then I'll also make a caddy-dns repo if you happen to be a Caddy user and would like to use this with Caddy. Thanks!

PR #112, opened 2023-12-18
Provider: HTTPReq

### LINKED ISSUE #111: Provider: acmeproxy / httpreq
You get to decide what we should call it here.

lego calls it httpreq
acme.sh calls it acmeproxy

But I would like to make the provider :)

KalleDK:
I've implemented the HTTPReq here https://github.com/KalleDK/httpreq

closes #111


mholt:
The caddy module is very easy, mostly just copy+paste the boilerplate: https://github.com/caddy-dns/template -- and tweak a few things, and you're good to go!  :)

Issue #114, opened 2024-02-08
New Provider Bunny (bunny.net)


dennismuench:
Hi there,

I need bunny.net DNS support for caddy and created a corresponding provider using the template, while basing tests and structure on the [hetzner](https://github.com/libdns/hetzner) package. Tests are OK.

I prepared a repository for [transport](https://github.com/rewilok/libdnsbunny/tree/transfer)

```

BUNNY_TEST_API_KEY="***" BUNNY_TEST_ZONE="***.com" go test ./... -v
=== RUN   Test_AppendRecords
--- PASS: Test_AppendRecords (7.43s)
=== RUN   Test_DeleteRecords
--- PASS: Test_DeleteRecords (4.00s)
=== RUN   Test_GetRecords
--- PASS: Test_GetRecords (3.85s)
=== RUN   Test_SetRecords
--- PASS: Test_SetRecords (7.96s)
PASS
ok      github.com/rewilok/libdnsbunny  23.510s
```

mholt:
Thank you!

PR #115, opened 2024-02-08
New Provider Bunny (bunny.net) 

dennismuench:
Hi,

I hope this is the correct way.

Here is the corresponding issue: https://github.com/libdns/libdns/issues/114


mholt:
This looks like a mistake, it's just some newlines in the readme. The PR should contain the code so it can be reviewed (even though it won't be merged here).

That said, I found [your code](https://github.com/rewilok/libdnsbunny/tree/transfer) from a link in the other issue and gave it a quick once-over. It looks good AFAICT. A suggestion might be to properly encode a query string rather than doing string concatenation, but that's mainly it for now.

Thanks for contributing! I'll make a repo in libdns and caddy-dns for you. (caddy-dns will be for the Caddy plugin, if you are a Caddy user).

Issue #76, opened 2022-06-20
DNSimple Provider


git001:
I plan to add the https://github.com/dnsimple/dnsimple-go to libdns.

mholt:
Hey Aleks @git001 -- still have plans to do this?

git001:
Sorry not time this year, for that :sob:  
Should I close it and reopen when I have time for that?

mholt:
No, it's fine. I just wanted to get an update :)

monirz:
@mholt Since it is not implemented yet or no one seems to be actively working on it yet. I'll be working on implementing it.

mholt:
Awesome!! Thank you so much @omarjatoi. I'm sure many people will find your contribution useful :)

mholt:
Huh, that's weird. You definitely have maintainer permissions. Make sure you're pushing with a key/login associated with your GitHub account...

mholt:
Hmmm... weird! Ok keep me posted if I need to do anything else.

PR #121, opened 2024-05-08
Add DNSimple implementation

omarjatoi:
Implementation for DNSimple (https://github.com/libdns/libdns/issues/76)


dnsimple/provider.go
```diff
@@ -0,0 +1,296 @@
...
+			deleted = append(deleted, r)
+		case http.StatusBadRequest:
+			fmt.Printf("Received a HTTP 400, could not delete record: %s", r.Name)
+			failed = append(failed, r)
+		case http.StatusUnauthorized:
+			fmt.Printf("Received a HTTP 401, suggesting authentication issues with the DNSimple client")
+			failed = append(failed, r)
+		default:
+			failed = append(failed, r)
+		}
```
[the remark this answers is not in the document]
mholt: If it were me, I'd probably just return after the first error, as the caller will need to rectify something and then try again. If an error is returned I think it's expected that the state may be unstable.

dnsimple/provider.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	// If we received records without an ID earlier, we're going to try and figure out the ID by calling
+	// GetRecords and comparing the record name. If we're able to find it, we'll delete it, otherwise
+	// we'll append it to our list of failed to delete records.
+	if len(noID) > 0 {
+		fetchedRecords, err := p.GetRecords(ctx, zone)
+		if err != nil {
+			fmt.Printf("Failed to populate IDs for records where one wasn't provided, err: %s", err.Error())
```
mholt, line 247: Printing to stdout is probably not _ideal_ but this isn't a showstopper.

dnsimple/provider.go
```diff
@@ -0,0 +1,296 @@
...
+						}
+						// See https://developer.dnsimple.com/v2/zones/records/#deleteZoneRecord for API response codes
+						switch resp.HTTPResponse.StatusCode {
+						case http.StatusNoContent:
+							deleted = append(deleted, r)
+						case http.StatusBadRequest:
+							fmt.Printf("Received a HTTP 400, could not delete record: %s", r.Name)
+							failed = append(failed, r)
+						case http.StatusUnauthorized:
+							fmt.Printf("Received a HTTP 401, suggesting authentication issues with the DNSimple client")
```
mholt, line 269: Maybe use `log.Printf` instead...

mholt (approved):
Thanks for contributing this!

I'll close this and create the libdns/dnsimple repo for you to push to, as well as a Caddy one in case you use Caddy, for the DNS provider plugin.

PR #125, opened 2024-07-09
Add do.de provider

### LINKED ISSUE #122: New Provider: do.de
I am not sure how to implement it. do.de only offers an `https://www.do.de/api/letsencrypt` endpoint that allows creating and deleting TXT records. There is no public API to manipulate other records.

[API Documentation](https://web.archive.org/web/20210307015600/https://www.do.de/wiki/LetsEncrypt_-_Entwickler) (archived, but the API is still the same)

d-k-bo:
This is a partial implementation of the libdns interfaces for [do.de](https://do.de). Since the do.de API only supports creating TXT records for ACME, an error is returned in any other situation.

This is my first time using Go, so don't be afraid to point out any style issues, errors, bad practices etc.

Closes #122

### DIFF
--- a/dode/client.go
+++ b/dode/client.go
@@ -1,0 +1,95 @@
+// Direct implementation of all required do.de DNS API endpoints
+
+package dode
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+const baseURl = "https://www.do.de/api"
+
+type ApiResponse struct {
+	Domain  *string `json:"domain"`
+	Success *bool   `json:"success"`
+	Error   *string `json:"error"`
+}
+
+func (p *Provider) createACMERecord(ctx context.Context, domain string, value string) error {
+	baseURL, _ := url.Parse(baseURl)
+	endpoint := baseURL.JoinPath("letsencrypt")
+
+	query := endpoint.Query()
+	query.Set("token", p.APIToken)
+	query.Set("domain", domain)
+	query.Set("value", value)
+
+	endpoint.RawQuery = query.Encode()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, endpoint.String(), http.NoBody)
+	if err != nil {
+		return err
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+
+	var apiResponse ApiResponse
+	decoder := json.NewDecoder(resp.Body)
+	err = decoder.Decode(&apiResponse)
+	if err != nil {
+		return err
+	}
+
+	if apiResponse.Error != nil {
+		return fmt.Errorf(*apiResponse.Error)
+	}
+	if apiResponse.Success == nil || !*apiResponse.Success {
+		return fmt.Errorf("creating the ACME record was not successfull")
+	}
+
+	return nil
+}
+
+func (p *Provider) deleteACMERecord(ctx context.Context, domain string) error {
+	baseURL, _ := url.Parse(baseURl)
+	endpoint := baseURL.JoinPath("letsencrypt")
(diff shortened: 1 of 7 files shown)


dode/client.go
```diff
@@ -0,0 +1,95 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+const baseURl = "https://www.do.de/api"
+
+type ApiResponse struct {
```
mholt, line 15:
```suggestion
type APIResponse struct {
```

This is a very unimportant nit, but in Go, convention is to typically capitalize initialisms like API, DNS, etc. etc.

dode/client.go
```diff
@@ -0,0 +1,95 @@
...
+	endpoint := baseURL.JoinPath("letsencrypt")
+
+	query := endpoint.Query()
+	query.Set("token", p.APIToken)
+	query.Set("domain", domain)
+	query.Set("value", value)
+
+	endpoint.RawQuery = query.Encode()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, endpoint.String(), http.NoBody)
```
mholt, line 32: TIL about `http.NoBody`. :+1:

dode/client.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+
+	var apiResponse ApiResponse
+	decoder := json.NewDecoder(resp.Body)
+	err = decoder.Decode(&apiResponse)
```
mholt, line 44:
```suggestion
	err = json.NewDecoder(resp.Body).Decode(&apiResponse)
```

One line :smiley:

dode/provider.go
```diff
@@ -0,0 +1,82 @@
...
+
+}
+
+// SetRecords sets the records in the zone, either by updating existing records or creating new ones.
+// It returns the updated records.
+//
+// The do.de API only supports creating TXT records that start with `_acme-challenge.`.
+func (p *Provider) SetRecords(ctx context.Context, zone string, records []libdns.Record) ([]libdns.Record, error) {
+	return p.AppendRecords(ctx, zone, records)
+}
```
mholt, line 56: Probably remove this, since Set doesn't actually set, it Appends. (Append is what is needed for ACME challenges.)

dode/provider.go
```diff
@@ -0,0 +1,82 @@
...
+
+const notSupportedErrorMsg = "the do.de API only supports creating and removing TXT records for domains starting with '_acme-challenge.'"
+const acmeChallenge = "_acme-challenge."
+
+// GetRecords lists all the records in the zone.
+//
+// This is unsupported by the do.de API.
+func (p *Provider) GetRecords(ctx context.Context, zone string) ([]libdns.Record, error) {
+	return nil, fmt.Errorf(notSupportedErrorMsg)
+}
```
mholt, line 28: You can remove this since it doesn't do anything.

dode/client.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return err
+	}
+
+	if apiResponse.Error != nil {
+		return fmt.Errorf(*apiResponse.Error)
```
mholt, line 88:
```suggestion
		return fmt.Errorf("API error: %v", *apiResponse.Error)
```

Usually it's a good idea to prefix errors with where it's coming from so you can trace it down easier.

mholt (commented):
Looks good overall, nice work especially for a first Go project! I'll create repos for you to push this to, and the corresponding caddy-dns module if you decide to do that later on so it can be used in Caddy.

I just left some comments that aren't critical.

d-k-bo:
Thanks!

Issue #126, opened 2024-07-23
LuaDNS provider


vitalie:
Hi,

I would like to add LuaDNS provider to libdns using the [Go client](https://github.com/luadns/luadns-go).

I've done some work already here:

https://github.com/luadns/libdns

If it looks good to you, I'll make a pull request to https://github.com/libdns/libdns.

Thanks!

mholt:
Thanks again for contributing this!

PR #127, opened 2024-07-26
Add LuaDNS provider

vitalie:
Hi,

This is implementation of [LuaDNS](https://www.luadns.com) provider, see [#126](https://github.com/libdns/libdns/issues/126).


mholt (approved):
Great! Thanks for the contribution. I'll create the repo for you to push to, as well as a caddy-dns repo in case you want to use it with Caddy.

PR #130, opened 2024-09-07
Add Nanelo DNS provider

AnTheMaker:
Hi there!
I've created a libdns (and caddy-dns) implementation for [Nanelo DNS](https://nanelo.com)!

libdns implementation: https://github.com/tibush/libdns-nanelo
caddy-dns implementation: https://github.com/tibush/caddy-dns-nanelo


mholt (approved):
Thanks for the contribution! It might be worth making a note in the readme or something that this only supports Append and Delete. That's totally fine but just might be worthwhile in case people use this for something other than ACME challenges later.

AnTheMaker:
Will do, thank you!

PR #132, opened 2024-09-08
Selectel v2

jjazzme:
new provider for [Selectel DNS v2](https://developers.selectel.ru/docs/cloud-services/dns_api/dns_api_actual/)


mholt:
Thanks for the contribution!

Overall this looks good, just wondering about the name. Generally I wouldn't want to have a separate repo per version of an API. Can we just name the package `selectel`? Instead of `selectelv2`. Thanks!

jjazzme:
> Thanks for the contribution!
> 
> Overall this looks good, just wondering about the name. Generally I wouldn't want to have a separate repo per version of an API. Can we just name the package `selectel`? Instead of `selectelv2`. Thanks!

There is a slight issue at the moment, as both versions of the API - v1 and v2 - are currently operational. They are tied to different DNS services of this provider. The provider says they will gradually discontinue support for v1, but apparently, it will remain a working version of the API for at least six more months. Is that alright? Shall we rename it?

jjazzme:
image

jjazzme:
@mholt, strategically, it's probably indeed better to rename it to selectel. I think v1 won't be around in a year. And if someone urgently needs libdns for v1, they can use name "selectel_v1". Okay, I'll rename it.

mholt:
Okay, yeah, I think we should just call it `selectel` and maybe mention in the README that only v2 API is supported. :+1:

I'll make the repos for you!

jjazzme:
@mholt  ok, but

jjazzme@Mac-Pro-jjazzme selectel % git tag -a "v1.0.0" -m "v1.0.0"
jjazzme@Mac-Pro-jjazzme selectel % git push --tags
remote: Permission to libdns/selectel.git denied to jjazzme.
fatal: unable to access 'https://github.com/libdns/selectel.git/': The requested URL returned error: 403

mholt:
@jjazzme It looks like you need to accept the invite to maintain that repo

Issue #133, opened 2024-09-24
Add Exoscale DNS Provider


mlec1:
Hello, 

I would like to add [Exoscale](https://www.exoscale.com/) as a DNS Provider to libdns. 

I have created this [repository](https://github.com/mlec1/libdns-exoscale) and will link a merge request

PR #134, opened 2024-09-24
Add Exoscale Provider

mlec1:
Related ticket is [here](https://github.com/libdns/libdns/issues/133)


mholt:
Great, thanks! I'll give you push access, and also create a caddy-dns repo for it in case you happen to use this with Caddy.

PR #144, opened 2024-12-28
DNSExit implementation

bouncysteve:


models.go
```diff
@@ -0,0 +1,35 @@
+package dnsexit
+
+type Action int64
+
+const (
+	Set Action = iota
+	Append
+	Delete
+)
```
mholt, line 9: You can probably (?) get away with un-exporting these (lowercase). But not a dealbreaker, just a suggestion.

mholt (approved):
Nice job for your first contribution in Go! LGTM, I will make a repo for libdns and another for caddy-dns which is for use in Caddy, if interested. I just ask that you maintain the package(s). Thank you!

bouncysteve:
Thanks for the review and the repos. Happy New Year!

PR #151, opened 2025-03-08
Implement provider for domainname.shop

JMyklebust:
Hi, I've made a provider for use with the Norwegian registrar/DNS provider domainname.shop (alt. name: domene.shop).
Not really a Go developer, but I've used the Hetzner DNS module as a basis and borrowed a lot from that module (though tried to rethink some things along the way if I could).

### DIFF
--- a/domainnameshop/LICENSE
+++ b/domainnameshop/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025 JMyklebust
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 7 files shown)


mholt:
Thanks for the contribution! Overall, at a quick glance, it looks good. I assume you've tested it and such. You'll be the maintainer of the repo(s) going forward.

I'll make the repos and give you push access. One for libdns, and another for a Caddy module in case that's relevant.

JMyklebust:
Thanks! Yeah I've tested reasonably well, I've been working on the Caddy bit as well so thanks for creating the repo there too.
That extra commit there was because I discovered that using Caddy, I got records that was getting the domainname.shop's default TTL of 1 hour which is not ideal. 😅
Didn't notice since the provider test specifies TTL so it's not a problem there (and the test data is constant too).

Issue #12, opened 2020-09-12
Clarify libdns API regarding DNS records names being absolute or relative


delthas:
Hi,

It is unclear whether the DNS record names (`libdns.Record.Name`):
1) use the standard "dot-suffix" relative/absolute notation (eg if zone FQDN is `example.org.` then `subdomain.example.org.` -> `subdomain.example.org` ; `subdomain` -> `subdomain.example.org`)
2) always use absolute paths implicitly (`subdomain.example.org` -> `subdomain.example.org`)

I think 1) is better as it directly reflects what a DNS server actually returns, but it seems that currently the API contract is implicitly 2).

Examples:
- the [gandi provider](https://github.com/libdns/gandi/blob/96971acb8aba44b770a309b0c21359a566dda369/client.go#L20) has to manually trim the `.example.org` part to get a relative notation to pass that to Gandi (notice how it expects to receive `subdomain.example.org` without an ending dot suffix) which uses the standard dot-suffix notation
- the [dnspod provider](https://github.com/libdns/dnspod/blob/97087b344e0abeb02eaba0f59c5fadc8dab92381/client.go#L97) behaves the same (but is buggier because it replaces all instances of the zone in the domain name which is incorrect)
- the cloudflare provider passes the value as-is to the API, but the API docs clearly show that the API uses [absolue paths implicitly](https://api.cloudflare.com/#dns-records-for-a-zone-create-dns-record) (see `example.org`, should be `example.org.` if using the dot-suffix notation, but here's it's just `example.org`)
- the [digitalocean provider](https://github.com/libdns/digitalocean/blob/f11d70f2506c1a0a9dad574e42d3767e14cb4a39/client.go#L42) [*does not work*](https://github.com/libdns/digitalocean/pull/3) becauses libdns passes implicitly absolute paths without a dot suffix and the provider passes it as is to the API which expects the dot-suffix notation (`1)`)

I suggest to choose which syntax/semantics we want to use for DNS record names, document it clearly and update the various libdns providers as necessary. I think using the standard absolute/relative notation depending on a final dot is better as it better correponds to what a DNS server returns.

mholt:
Thanks for bringing this up. I think I had noticed this somewhere along the way but lost track of it before I did anything about it because of everything else going on recently.

[...]

PR #152, opened 2025-03-09
Document exact semantics of DNS interfaces

### LINKED ISSUE #145: Clarify semantics for SetRecords, GetRecords and Close
The libdns providers made it possible for me to make https://github.com/mjl-/dnsclay, providing a standard DNS UPDATE/AXFR/NOTIFY interface for all these providers with their custom APIs. Great to have so many! I ran into some questions about the semantics of the Records methods.

For SetRecords, I wonder about the semantics with resource record sets with more than 1 already existing record. Should provider implementations be grouping their records parameter into RR sets, and then replacing the entire existing set, deleting any other existing records in those sets that are no longer mentioned? Perhaps it makes sense to only specify behaviour for providers with IDs for records: SetRecords would update just those records. Callers can then use DeleteRecords and AppendRecords to remove from/add records to a record set. It could be good to mention the concept of resource record set in the docs, so implementers are aware of those semantics, provider APIs may need special handling (like with the route53 provider).

For GetRecords, are SOA and DNSSEC-related records like RRSIG, NSEC, NSEC3 supposed to be returned? I ran into a few that don't return the SOA record, which is likely the fault of the DNS operator API. For DNSSEC, some DNS operators will generate those records on demand, and won't have them stored in the zone, while others will, so in practice I suspect this will be optional. May be good to document so consumers know what to expect.

Perhaps a Provider should have an (optional?) Close method? So e.g. http.Clients with custom http.Transports (directadmin, inwx) can be closed?

gucci-on-fleek:
I'm posting these documentation suggestions as a draft PR, as suggested in https://github.com/libdns/libdns/issues/145#issuecomment-2708503004. I'll discuss the specific changes below, but some more general comments:

1. I've generally tried to document this “how they work right now” rather than “how they should work”, although I haven't actually looked through very many provider implementations, so some of my assumptions may be incorrect.
2. I have very little Go experience, so I'm not very familiar with the Go documentation conventions.
3. I'm just a hobbyist and certainly not a DNS expert, so if something looks incorrect, then it's probably me that made the mistake.

Thanks!

Closes #145.

### DIFF
--- a/libdns.go
+++ b/libdns.go
@@ -1,41 +1,41 @@
-// Package libdns defines core interfaces that should be implemented by DNS
+// Package [libdns] defines core interfaces that should be implemented by DNS
 // provider clients. They are small and idiomatic Go interfaces with
-// well-defined semantics for the purposes of reading and manipulating
-// DNS records using DNS provider APIs.
+// well-defined semantics for the purposes of reading and manipulating DNS
+// records using DNS provider APIs.
 //
-// This documentation uses the definitions for terms from RFC 7719:
+// This documentation uses the definitions for terms from [RFC 7719].
 //
-//	https://datatracker.ietf.org/doc/html/rfc7719
+// Records are described independently of any particular zone, a convention that
+// grants [Record] structs portability across zones. As such, record names are
+// partially qualified, i.e. relative to the zone. For example, an “A” record
+// called “sub” in zone “example.com.” represents a fully-qualified domain name
+// (FQDN) of “sub.example.com.”. Implementations should expect that input
+// records conform to this standard, while also ensuring that output records do;
+// adjustments to record names may need to be made before or after provider API
+// calls, for example, to maintain consistency with all other [libdns] packages.
+// Helper functions are available in this package to convert between relative
+// and absolute names.
 //
-// Records are described independently of any particular zone, a convention
-// that grants Record structs portability across zones. As such, record names
-// are partially qualified, i.e. relative to the zone. For example, an A
-// record called "sub" in zone "example.com." represents a fully-qualified
-// domain name (FQDN) of "sub.example.com.". Implementations should expect
-// that input records conform to this standard, while also ensuring that
-// output records do; adjustments to record names may need to be made before
-// or after provider API calls, for example, to maintain consistency with
-// all other libdns packages. Helper functions are available in this package
-// to convert between relative and absolute names.
-//
-// Although zone names are a required input, libdns does not coerce any
+// Although zone names are a required input, [libdns] does not coerce any
 // particular representation of DNS zones; only records. Since zone name and
-// records are separate inputs in libdns interfaces, it is up to the caller
-// to pair a zone's name with its records in a way that works for them.
+// records are separate inputs in [libdns] interfaces, it is up to the caller to
+// pair a zone's name with its records in a way that works for them.
 //
 // All interface implementations must be safe for concurrent/parallel use,
-// meaning 1) no data races, and 2) simultaneous method calls must result
-// in either both their expected outcomes or an error.
+// meaning 1) no data races, and 2) simultaneous method calls must result in
+// either both their expected outcomes or an error.
 //
-// For example, if AppendRecords() is called at the same time and two API
-// requests are made to the provider at the same time, the result of both
-// requests must be visible after they both complete; if the provider does
-// not synchronize the writing of the zone file and one request overwrites
-// the other, then the client implementation must take care to synchronize
-// on behalf of the incompetent provider. This synchronization need not be
-// global; for example: the scope of synchronization might only need to be
-// within the same zone, allowing multiple requests at once as long as all
-// of them are for different zones. (Exact logic depends on the provider.)
+// For example, if [libdns.RecordAppender.AppendRecords] is called at the same
+// time and two API requests are made to the provider at the same time, the
+// result of both requests must be visible after they both complete; if the
+// provider does not synchronize the writing of the zone file and one request
+// overwrites the other, then the client implementation must take care to
+// synchronize on behalf of the incompetent provider. This synchronization need
+// not be global; for example: the scope of synchronization might only need to
+// be within the same zone, allowing multiple requests at once as long as all of
+// them are for different zones. (Exact logic depends on the provider.)
+//
+// [RFC 7719]: https://datatracker.ietf.org/doc/html/rfc7719
 package libdns
 
 import (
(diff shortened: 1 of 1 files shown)


libdns.go
```diff
@@ -43,7 +47,8 @@ import (
 
 // RecordGetter can get records from a DNS zone.
 type RecordGetter interface {
-	// GetRecords returns all the records in the DNS zone.
+	// GetRecords returns all the records in the DNS zone. DNSSEC-specific
+	// records may or may not be included in an implementation-defined way.
```
gucci-on-fleek, line 51: Rationale: Most providers implement DNSSEC automatically (or not at all) rather than expecting users to manually sign their zone. Because of this, many providers don't include the DNSSEC records in their standard "get zone" API calls, even though they're technically just standard records in the zone.

libdns.go
```diff
@@ -54,7 +59,8 @@ type RecordGetter interface {
 type RecordAppender interface {
 	// AppendRecords creates the requested records in the given zone
 	// and returns the populated records that were created. It never
-	// changes existing records.
+	// changes existing records. Therefore, it is invalid to use this
```
gucci-on-fleek, line 62: Rationale: Hopefully this makes it easier for people to decide whether to use `GetRecords` or `AppendRecords`.

libdns.go
```diff
@@ -69,25 +75,61 @@ type RecordSetter interface {
 	// maintain parity with the input. No other records are affected.
 	// It returns the records which were set.
 	//
-	// Records that have an ID associating it with a particular resource
-	// on the provider will be directly replaced. If no ID is given, this
-	// method may use what information is given to do lookups and will
-	// ensure that only necessary changes are made to the zone.
```
gucci-on-fleek, line 75: Rationale: We want for the interfaces to have consistent semantics between different providers, but not all providers give access to something like a "record ID", so there's no way for something like this to give consistent results.

libdns.go
```diff
@@ -69,25 +75,61 @@ type RecordSetter interface {
...
+	// For any (name, type) pair in the input, `SetRecords` ensures that the
+	// only records in the output zone with that (name, type) pair are those
+	// that were provided in the input.
+	//
+	// In RFC 7719 terms, `SetRecords` appends, modifies, or deletes records
+	// in the zone so that for each RRset in the input, the records provided
+	// in the input are the only members of their RRset in the output zone.
```
gucci-on-fleek, line 84: Rationale: This seems to be the semantics implied/expected by the [upstream Caddy support for ECH](https://github.com/caddyserver/caddy/blob/220cd1c2bcecc07bcf6a0141069538c1b1109907/modules/caddytls/ech.go#L665-L674), and seems to be the most reasonable definition to me.

One possible problem here is that if almost none of the providers currently use these semantics for `SetRecords`, then it might make more sense to define a new interface with these semantics and deprecate the current `SetRecords`.

libdns.go
```diff
@@ -69,25 +75,61 @@ type RecordSetter interface {
...
 
 // RecordDeleter can delete records from a DNS zone.
 type RecordDeleter interface {
-	// DeleteRecords deletes the given records from the zone if they exist.
-	// It returns the records that were deleted.
+	// DeleteRecords deletes the given records from the zone if they exist in
+	// the zone and exactly match the input. If the input records do not exist
+	// in the zone, they are silently ignored. `DeleteRecords` returns only the
+	// the records that were deleted, and does not return any records that were
+	// provided in the input but did not exist in the zone.
```
gucci-on-fleek, line 129: Rationale: I think that this is the semantics intended by the current interfaces? But it wasn't obvious before if trying to delete a non-existent record would lead to an error or not.

libdns.go
```diff
@@ -69,25 +75,61 @@ type RecordSetter interface {
...
 	//
-	// Records that have an ID to associate it with a particular resource on
-	// the provider will be directly deleted. If no ID is given, this method
-	// may use what information is given to do lookups and delete only
-	// matching records.
+	// Note that it is semantically invalid to remove the last NS record from a
+	// zone, so attempting to do is undefined behavior.
```
gucci-on-fleek, line 132: I'm not sure if this is helpful context or irrelevant trivia.

libdns.go
```diff
@@ -105,26 +147,87 @@ type ZoneLister interface {
 }
 
 // Record is a generalized representation of a DNS record.
-//
-// The values of this struct should be free of zone-file-specific syntax,
-// except if this struct's fields do not sufficiently represent all the
-// fields of a certain record type; in that case, the remaining data for
-// which there are not specific fields should be stored in the Value as
-// it appears in the zone file.
```
gucci-on-fleek, line 113: Rationale: It seems like most of the RR types _do_ expect a value encoded using zone-file-specific syntax, with only a few special cases that use other struct fields.

libdns.go
```diff
@@ -105,26 +147,87 @@ type ZoneLister interface {
...
+	// or may not support any given record type, and may support additional
+	// "private" record types with implementation-defined behavior.
+	//
+	// Examples: "A", "AAAA", "CNAME", "MX", "TXT"
+	Type string
+
+	// The `Name` field specifies the name of the record. It is partially
+	// qualified relative to the current zone. This field is called a "Label" by
+	// RFC7719. You may use "@" or the empty string "" to represent the root of
+	// the zone.
```
gucci-on-fleek, line 166: > You may use "@" or the empty string "" to represent the root of the zone.

I'm not actually sure how you are currently expected to specify the root of the current zone, so this is just my best guess at what should work.

libdns.go
```diff
@@ -105,26 +147,87 @@ type ZoneLister interface {
...
+	//
+	// Examples: "www", "", "@", "subdomain", "sub.subdomain"
+	//
+	// Invalid: "www.example.com.", "example.net." (fully-qualified)
```
gucci-on-fleek, line 170: The original comment said

> partially-qualified (relative to zone)

but it might be useful to allow using a FQDN here (with the trailing dot).

libdns.go
```diff
@@ -105,26 +147,87 @@ type ZoneLister interface {
...
+	//           (SVCB)  10
+	Priority uint
+
+	// The `Weight` field specifies the weight of the record. This field is
+	// only applicable for certain record types, but is mandatory for those
+	// types.
+	//
+	// Examples: (SRV) 20
+	//           (URI) 20
+	Weight uint
```
gucci-on-fleek, line 221: Considering that only `SRV` and `URI` records use the `Weight` field, the `SRV` records already force you to encode the port inside the `Value` field, and essentially nobody uses `URI` records, I wonder if it might make more sense to deprecate this field and recommend just encoding inside of the `Value` field. It's not great to need to encode the integer weight inside of the string `Value`, but you already have to do so for the port, so it shouldn't make much of a difference for the `libdns` interface users.

libdns.go
```diff
@@ -105,26 +147,87 @@ type ZoneLister interface {
...
+
+	// The `Target` field specifies the target of the record. This field is
+	// only valid for HTTPS and SVCB records, and *not* for SRV, MX, or CNAME
+	// records, which store their targets in the `Value` field. This field must
+	// be set to the fully-qualified domain name (FQDN) of the target.
+	//
+	// Examples: (HTTPS) "example.com."
+	//           (SVCB)  "example.com."
+	Target string
```
gucci-on-fleek, line 230: The original comment here suggested that the `Target` field is only valid for `HTTPS` records (and `SVCB` records too, presumably). This is fairly confusing though since `SRV`, `MX`, and `CNAME` records all have a field traditionally called the "target".

libdns.go
mholt: Hm... yeah... the more I work with this the more I feel like Value should be the one and only "source of truth" for a record's value. I agree with this.

libdns.go
mholt: I am not sure either! Is it more conventional to use "" or "@"? Cloudflare, for instance, requires "@", but I don't know if that's a standard or just preference.

libdns.go
mholt: The main issue with allowing both relative and FQDN is, what do providers do in return? i.e. are Records returned by `GetRecords()`, for instance, filled with Relative or FQDN names? All libdns APIs are scoped by a particular zone, so making the names be relative keeps them portable, which seems useful. It _does_ imply that zone information needs to be carried along with records, which I can understand may be inconvenient, but I think most DNS provider APIs expect zone-relative names here because most APIs also seem to operate within the scope of specific zones.

So... sure, allowing either might simplify application code, but I could also see errors here (like, "oops, forgot the trailing dot, since nobody actually uses those except in DNS zone files"). But then again, we can't completely abstract away DNS zone file syntax (see the discussion over the Value field).

Sigh... I dunno. My gut still tells me that relative-only is best, but I'm open to discussing this! See previous discussion here: https://github.com/libdns/libdns/issues/12

libdns.go
```diff
@@ -105,26 +147,87 @@ type ZoneLister interface {
...
+	//                   Also note that the target is included here, and not
+	//                   in the `Target` field.)
+	//	         (HTTPS) "alpn=h2,h3 port=443" (Note that this excludes the
+	//                   priority field and target fields.)
 	Value string
-	TTL   time.Duration
+
+	// The `TTL` field specifies the time-to-live of the record. This is
+	// represented in the DNS as an unsigned integral number of seconds, but
+	// is provided here as a time.Duration.
```
mholt, line 198: Maybe we should mention that fractional seconds are truncated?

libdns.go
mholt: I'm actually starting to agree with you on this. We can provide helper methods to parse out type-specific fields from the Value, but I am really starting to think that the Value should remain an opaque string...

mholt (changes requested):
This is a great start! Thanks for taking initiative on this.

Some difficult questions to answer, for sure. Let's keep hashing this out!

I like the use of examples, nicely done on those.

libdns.go
mholt: Oops, that's true -- I only mentioned HTTPS records in the initial comment since that's what I used them for.

I'm actually rethinking these extra fields.

To elaborate from other comments, I wonder if Record should stick to _only_ a generic representation of a DNS record, with opaque Value. And then for type-specific fields we use specific struct types with the parsed values in separate fields. So like, a `type SRV struct` or `type HTTPS struct`, etc, with helper methods to go between the two.

mholt (changes requested):
Thanks for the discussion! I think we're making progress. Curious to know your thoughts.

I don't expect you to design a bunch of new struct types, but I think we are starting to converge on a plan for a stable 1.0 libdns API.

libdns.go
gucci-on-fleek: > Oof, sorry, I meant the semantics you are suggesting in this change.

:+1:

> > How is Caddy supposed to know which HTTPS record to update? I think that the easiest answer to this is to refuse to set the ech= key unless there is only 1 HTTPS record for the initial implementation, and wait to see if there's user demand for multiple HTTPS records. And if you're only using one HTTPS record, then it doesn't really matter what priority you set, so you might as well use 1.
> 
> Dang, also a really good point. I agree. We should reject auto-setting HTTPS records if there is more than one priority-2 HTTPS record with an "ech" key. (Did I get that right? If so, I'll implement that change in Caddy right away.)

I think that only the following cases should be accepted by Caddy:

1. ((Exactly 1 `HTTPS` record) with any contents _except_ for `priority == 0`). This should always be safe to add the `ech=` value to.
2. ((Exactly 1 `HTTPS` record with (`priority == 2` and the `ech=` value set)) and (any number of `HTTPS` records with (`priority != 2` or `ech == nil`))). This sorta implies that the other `HTTPS` records are “aware” of Caddy and are okay coexisting.

I think that these are the only unambiguous cases, but there might be a few more cases that are still safe.

libdns.go
gucci-on-fleek: > I'm actually rethinking these extra fields.
> 
> To elaborate from other comments, I wonder if Record should stick to _only_ a generic representation of a DNS record, with opaque Value. And then for type-specific fields we use specific struct types with the parsed values in separate fields. So like, a `type SRV struct` or `type HTTPS struct`, etc, with helper methods to go between the two.

I agree. Probably the easiest option would be to use the [structs defined by `github.com/miekg/dns`](https://pkg.go.dev/github.com/miekg/dns#pkg-types), which is what [`libdns/rfc2136`](https://github.com/libdns/rfc2136/blob/master/conversion.go) and DNSControl ([1](https://github.com/StackExchange/dnscontrol/blob/5dafc9c1e3caaa4c3142e3009aa90db43f11c9ce/models/dnsrr.go), [2](https://github.com/StackExchange/dnscontrol/blob/5dafc9c1e3caaa4c3142e3009aa90db43f11c9ce/models/record.go)) both do already. That interface is a little bit annoying though since instead of using a `Value` field on the structs, for `A` RRs it uses a `A` field, for `AAAA` RRs it uses a `AAAA` field, for `CNAME` RRs it uses a `Target` field, etc.

libdns.go
```diff
@@ -219,10 +349,15 @@ func RelativeName(fqdn, zone string) string {
...
+		// allow for that. I guess return an empty string?
+		return ""
+	}
 	if zone == "" {
 		return strings.Trim(name, ".")
 	}
-	if name == "" || name == "@" {
+	if name == "@" {
 		return zone
 	}
```
gucci-on-fleek, line 362: Should I change this function signature to return an error if `""` is passed, or just return the empty string `""` as I am right now, or something else?

libdns.go
mholt: I was looking at miekg/dns, useful library -- but maybe a bit too low-level for libdns? I also want to avoid extra dependencies when possible (and miekg/dns, while useful, is quite a large package).

What about this for our package:

```go
type Record struct {
	// provider-specific metadata
	ID string

	// general record fields
	Type  string
	Name  string // partially-qualified (relative to zone)
	Value string // opaque string
	TTL   time.Duration
}

type HTTPSRecord struct {
    Record
    Target    string
    Priority  uint16
    SvcParams map[string][]string
}
```

i.e. embedding Record in every type-specific struct. Or something like this.

And then we can have methods to go between Record and type-specific structs.

libdns.go
mholt: Isn't it okay if the name is empty string? It just means the FQDN is the zone, right?

mholt (changes requested):
Great discussion!! We're getting close.

libdns.go
```diff
@@ -2,6 +2,10 @@
 // provider clients. They are small and idiomatic Go interfaces with
 // well-defined semantics.
 //
+// Note: This documentation uses the definitions for terms from RFC 7719:
+//
+//	https://datatracker.ietf.org/doc/html/rfc7719
+//
```
mholt, line 8:
```suggestion
// provider clients. They are small and idiomatic Go interfaces with
// well-defined semantics for the purposes of reading and manipulating
// DNS records using DNS provider APIs.
//
// This documentation uses the definitions for terms from RFC 7719:
//
//	https://datatracker.ietf.org/doc/html/rfc7719
//
```

libdns.go
gucci-on-fleek: > What if records with IDs were rejected with an error if the provider does not support IDs?

The problem that I was suggesting was that if the maintainer of a module that uses libdns only ever tested it with ID-supporting providers, then the module might inadvertently delete records for non-ID-supporting providers. This would be easily noticed if the module maintainer tested with multiple providers, but most people only have accounts with 1 or 2 DNS providers at most, so this isn't that easy to do.

Or maybe I'm just misunderstanding something here? I haven't really used any of the libdns APIs myself, so this is quite likely.

> Obviously, IDs are the ideal way to manage records, so it'd be nice to make it possible to use them if available, 

Agreed. I wonder if it's possible to always synthesize an ID? Like, if the IDs are just an opaque string, could the ID just fallback to being the complete DNS record as a zone-file-formatted string?

libdns.go
gucci-on-fleek: > I think I follow that reasoning, and I do think it makes sense. But given the complexity, what if I just start with:
> 
> _Only set the HTTPS record if there isn't any, OR (if there is only one AND (it does not have an `ech` key OR `priority != 2`))._
> 
> That's much simpler and should cover 99.9% of use cases for now, I suppose, until we get some wider use.

Yeah, that seems fine to me.

(Practically speaking, I think that pretty much anything will be fine. Users doing normal things will either have 0 or 1 `HTTPS` records, in which case Caddy will do the right thing. And any users who have multiple `HTTPS` records are doing *weird DNS stuff*™, in which case they should be able to debug any issues themselves.)

libdns.go
gucci-on-fleek: > I was looking at miekg/dns, useful library -- but maybe a bit too low-level for libdns? I also want to avoid extra dependencies when possible (and miekg/dns, while useful, is quite a large package).
> 
> What about this for our package:
> 
> [...]
> 
> i.e. embedding Record in every type-specific struct. Or something like this.
> 
> And then we can have methods to go between Record and type-specific structs.

I guess the question is, how general of a package do you want this to be. If you only want to support RR types that people are actually to use, then this should be fine; there are only a dozen common RR types, only half of which have special fields. But if you want to support most/all RR types, then this will be a lot of work; take a look at [`AMTRELAY`](https://pkg.go.dev/github.com/miekg/dns#AMTRELAY) for example.

I think that only supporting the somewhat-common types is perfectly reasonable, but it really depends on what your goals are here.

libdns.go
gucci-on-fleek: That goes back to [`"@"` vs `""` discussion](https://github.com/libdns/libdns/pull/152#discussion_r1989557976), and it would seem weird to me to allow `""` here but forbid it in the `Record` struct. Or maybe these are different cases since the struct field gets initialized to `""` but you always need to explicitly pass in a function parameter?

libdns.go
mholt: Well, it definitely feels weird if that code doesn't work doesn't it... I feel like that should work.

What if the provider's DNSSEC behavior can be configured? Somewhere, there is a concrete type that the user has configured (with API key, etc). By default, all DNSSEC gets filtered out by all providers -- that way, calls always work reliably. But if a provider wanted to enable it, it could be configurable.

(If we do this, we should document what DNSSEC records look like. Maybe even provide some boilerplate code for them or an `if` statement or something. What would that look like?)

libdns.go
mholt: > The problem that I was suggesting was that if the maintainer of a module that uses libdns only ever tested it with ID-supporting providers, then the module might inadvertently delete records for non-ID-supporting providers. ...
>
> Or maybe I'm just misunderstanding something here?

No, I do think you're right.

What if we document that IDs should be stripped if using records across providers? We could even provide a helper method, I dunno.

And providers should reject records with IDs if they don't support record IDs, by returning an error.

I just think for the use cases where records have IDs, that will be invaluable. But yeah, for portability, we need to strip IDs, but that's an explicit "I am porting records across providers" which probably won't happen super often.

libdns.go
mholt: Sounds good to me!

libdns.go
mholt: Hmmmmmmm! No, that makes sense, I'm glad you've thought about it more carefully than I have.

If I get you right, you're basically saying: "If there is an ID, use that as the ID. If there is not an ID, use the RRset as an ID."

There may be unpleasant surprises if a record has an ID but the provider doesn't support them (or vice-versa: a record doesn't have an ID but the provider supports them), whereas all records are part of an RRset.

So... gosh, I dunno, maybe we drop the idea of an ID field?

I almost wonder if an "equivalence" type or function could be defined to help customize what equality means when comparing DNS records. But that would be a matter for another discussion I think (unless you have some ideas on that now).

mholt:
Alrighty, so, to wrap this PR up for now, do you want to make any final adjustments if you have any you want to do, and then I'll merge this, then I'll open a new PR with the discussed changes / expansion to the library? Would be happy to have your review at that point too.

libdns.go
gucci-on-fleek: > So... gosh, I dunno, maybe we drop the idea of an ID field?

Yeah, I think that that's the only fully-portable option here.

> I almost wonder if an "equivalence" type or function could be defined to help customize what equality means when comparing DNS records. But that would be a matter for another discussion I think (unless you have some ideas on that now).

Hmm, that's an interesting idea. I haven't thought about it too much though, so we can probably leave that for later.

gucci-on-fleek:
> Alrighty, so, to wrap this PR up for now, do you want to make any final adjustments if you have any you want to do,

Alright, done. The content should be the same as before, but I went ahead and reformatted it so that it displays better on https://pkg.go.dev. 

> I'll open a new PR with the discussed changes / expansion to the library? Would be happy to have your review at that point too.

Sure, sounds good. Please @mention me when you do.

Thanks again for your helpful reviews/comments!

mholt (approved):
Awesome! I'll get to work! (Might be after the weekend)

Issue #44, opened 2021-06-12
Unit tests for providers (and template)


rugk:
I've skimmed [across your implementation doc](https://github.com/libdns/libdns/wiki/Implementing-providers) and especially [your template repo](https://github.com/libdns/template), and I wonder whether some simple unit tests (included in the template) would not be a good idea?
Maybe also directly include a GitHub Action to run these tests in a CI.

As such, IMHO; this would make developing a new provider even easier by having a verification whether everything works (somewhat, at least) and support proper development practices.
This could make it easier to test the providers.

mholt:
Yeah, that'd be a good idea, although, it's not super clear to me what "unit tests" would look like, since the output of each function is an external side effect on the network. We'd probably need mocks, which have to be customized for each provider... I hesitate to do DNS lookups since we've found they're extremely finicky in CI and people's various dev machines because people have the most convoluted, weirdest DNS resolver configurations that break tests like this...

Probably the most useful tests would be those which verify that API requests are being formed correctly, rather than testing the actual execution of those requests. I don't know.

Adphi:
Conformance tests would work.
 
The main library implements the main conformance tests logic. This is what [Cert-Manager](https://github.com/cert-manager/cert-manager/blob/master/test/acme/dns/options.go) or the [CSI](https://github.com/kubernetes-csi/csi-test/tree/master/pkg/sanity) does.
The conformance tests only takes a `Provider` and each provider must ensure tests pass, so they validate it with their own development credentials.

mholt:
I like that idea. It'd be awesome if it could somehow get each provider to use a mock DNS system somehow, operated by the conformance tests. But I don't know if that's feasible.

Adphi:
While working with multiple implementations (scaleway, ovh, hetzner and cloudflare) what I found is that you can manage the zones in this providers without actually using their dns servers 😊. It just needs an account. I used one of my test domains for all of them but it may even work with anything, e.g. example.org

I think it should not use a mock at all as it would just validate the mock. Conformance tests should make real network calls with a test account and a test domain.

Adphi:
I know it is off topic but some kind of `Capabilities` interface to know what kind of records are supported by the provider could be very useful / helpful.

mholt:
Maybe, but I also kind of just expect DNS providers to support all the common record types, at least.

Adphi:
Exactly, you'd expect it. With capabilities, you'd be sure of it at runtime before any attempt.

[...]

PR #153, opened 2025-03-19
All-new Record abstraction and exported APIs

mholt:
Initial draft for v1.0 pre-releases.

Follow-up from #152.

This is a breaking change, but most provider implementations will remain mostly the same. Most will need to search-replace some identifiers and, with time, conform to the clarified semantics in some cases.

This change is crucial to solidify APIs for a permanent v1 release in the future, and to reduce the likelihood of breaking changes after that.

/cc @gucci-on-fleek -- thanks for spearheading this. I've tried to maintain the clarified semantics from the earlier PR.

The main change is:

`libdns.Record` is now an interface that simply returns an `RR`.

What used to be `Record` is now `RR`, and it is strictly a DNS record's main fields: Name, Type, TTL, and Data.

There are also type-specific structs such as `A`, `AAAA`, `CNAME`, `SRV`, etc., that implement `libdns.Record`.

To convert from `RR` to a specific type, there's `RR.Parse()`. To go the other way (from specific type to `RR`), there's the `RR()` method on every type (implementing the `Record` interface).

Since the main APIs (Get/Set/etc...) take the `Record` type, users have the option of using either RRs, or type-specific structs, in their application code, and provider packages can also choose which they prefer on their end.

Once we finalize these changes, we can attempt to notify provider package maintainers of the upcoming changes in the hopes for a stable API with 1.0 on the horizon. Since most packages are small, this shouldn't be too difficult. Also, many of the semantics are quite nuanced and some only pertain in edge cases, so it's okay if the semantics are implemented by each provider package gradually over time.

The main uses of this package -- that I know of -- are appending and deleting TXT records, updating A/AAAA records, and setting HTTPS records. I think they're mostly straightforward use cases that shouldn't surprise most users and package maintainers, except it might expose bugs that were previously less obvious.

Hopefully this is in the right direction! Let me know what you think.


mholt:
I've been updating the [cloudflare provider](https://github.com/libdns/cloudflare) in a branch to use this branch, to see how it goes. I'm testing with a SRV record currently... since it's more complicated.

Since the [Cloudflare documentation](https://developers.cloudflare.com/api/resources/dns/subresources/records/models/srv_record/#(schema) shows a "content" field and optional "data" components, I figured I can just pass in the content of the record as a single string and call it a day. And this code path, enabled by the changes in this PR, is MUCH simpler than we used to have before.

But, unfortunately, either their docs are wrong or their API is wrong: https://community.cloudflare.com/t/creating-srv-record-with-content-string-instead-of-individual-component-fields/781178?u=mholt -- I am getting errors when omitting the individual data components, despite the docs saying they are optional.

So, I'm hoping their API is just buggy and the docs are right, since otherwise I have to type-assert every single record type to use their API.

Anyway, this PR looks promising so far in my spike code, despite my disappointment if I end up having to type-assert each record type with Cloudflare...

gucci-on-fleek:
Looks good to me so far; I should have time for a more thorough review this weekend.

gucci-on-fleek (commented):
Seems excellent so far; thanks! Just a couple minor comments.

> What used to be `Record` is now `RR`, and it is strictly a DNS record's main fields: Name, Type, TTL, and Data.
>
> There are also type-specific structs such as `A`, `AAAA`, `CNAME`, `SRV`, etc., that implement `libdns.Record`.
>
> To convert from `RR` to a specific type, there's `RR.Parse()`. To go the other way (from specific type to `RR`), there's the `RR()` method on every type (implementing the `Record` interface).

:+1:

> Also, many of the semantics are quite nuanced and some only pertain in edge cases, so it's okay if the semantics are implemented by each provider package gradually over time.

Yeah, most of the recommended semantics probably don't need to be implemented until someone complains; at least with the semantics specified, it will be clear _where_ to make the fix (`libdns/<provider>`, `libdns/libdns`, `caddy-dns/provider`, `caddyserver/caddy`). Otherwise, I could see someone reporting something broken to a specific provider, it getting fixed there, then someone else reporting the same issue somewhere upstream, resulting in a double-fix which then breaks downstream again.

> I've been updating the [Cloudflare provider](https://github.com/libdns/cloudflare) in a branch that uses this branch. I'm testing with a SRV record currently, since it's a more complicated record type.
>
> So far, the effects of this refactor look promising.
>
> Except... something that doesn't really have much to do with this PR...
>
> Since the [Cloudflare documentation](https://developers.cloudflare.com/api/resources/dns/subresources/records/models/srv_record/#(schema)) shows a "content" field and optional "data" components, I figured I can just pass in the content of the record as a single string and call it a day. (That's easy now with the `RR` type.) It's MUCH simpler with this PR, than we had before.
>
> But, unfortunately, either their docs are wrong or their API is wrong: https://community.cloudflare.com/t/creating-srv-record-with-content-string-instead-of-individual-component-fields/781178?u=mholt -- I am getting errors when omitting the individual data components, despite the docs saying they are optional.
>
> So, I'm hoping their API is just buggy and the docs are right, since otherwise I have to type-assert every single record type to use their API. Doable, and inevitably some providers will require this, but kinda lame.

Ugh, that's disappointing. Hopefully not _all_ the providers are like that.

libdns.go
```diff
@@ -1,394 +1,194 @@
...
+// This package represents DNS records in two primary ways: as opaque [RR]
+// structs, where the data is serialized as a single string as in a zone file;
+// and as individual type structures, where the data is parsed into its separate
+// fields for easier manipulation by Go programs (for example: [SRV] and [HTTPS]
+// types). This hybrid design offers great flexibility for both DNS provider
+// packages and consumer Go programs.
+//
+// This package represents records flexibly with the [Record] interface, which
+// is any type that can transform itself into the [RR] struct, which is a
+// type-agnostic [Resource Record](https://en.wikipedia.org/wiki/Domain_Name_System#Resource_records)
```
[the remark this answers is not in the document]
mholt: D'oh, thanks. Good to know... will fix.

libdns.go
```diff
@@ -1,394 +1,194 @@
...
-// use.
 type RecordGetter interface {
+	// GetRecords returns all the records in the DNS zone.
+	//
+	// DNSSEC-related records are typically not included in the output, but this
+	// behavior is implementation-defined. If an implementation includes DNSSEC
+	// records in the output, this behavior should be documented.
+	//
+	// Implementations must honor context cancellation and be safe for concurrent
+	// use.
```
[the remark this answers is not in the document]
mholt: I see what you mean; I think I prefer the latter... the godoc still shows them, just in-line with the code, and then gopls does too.

record.go
```diff
@@ -0,0 +1,395 @@
...
+	// encouraged to support as many of the most common record types as possible,
+	// especially: A, AAAA, CNAME, TXT, HTTPS, and SRV.
+	//
+	// Other custom record types may be supported with implementation-defined
+	// behavior.
+	Type string
+
+	// The data (or "value") of the record. This field should be formatted in the
+	// standard zone file syntax (technically, the "RDATA" field as defined by
+	// RFC 1034).
```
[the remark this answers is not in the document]
mholt: Good thinking... I think I agree on all points.

record.go
```diff
@@ -0,0 +1,395 @@
...
+		Name:   r.Name,
+		TTL:    r.TTL,
+		Target: r.Data,
+	}, nil
+}
+
+func (r RR) toHTTPS() (HTTPS, error) {
+	if expectedType := "HTTPS"; r.Type != expectedType {
+		return HTTPS{}, fmt.Errorf("record type not %s: %s", expectedType, r.Type)
+	}
```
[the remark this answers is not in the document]
mholt: Ugh, yeah, that's true...

This is how the miekg/dns package does it: https://pkg.go.dev/github.com/miekg/dns#HTTPS

I might take a cue from that. Will (probably) update this...

rrtypes.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func (aaaa AAAA) RR() (RR, error) {
+	return RR{
+		Name: aaaa.Name,
+		TTL:  aaaa.TTL,
+		Type: "AAAA",
+		Data: aaaa.IP.String(),
+	}, nil
+}
```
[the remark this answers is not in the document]
mholt: Yeah, good idea... I guess. Feels a little weird to break the pattern though and have a RR type that isn't named the same.

Are there any other semantic differences between the two record types that you know of? If not... then yeah, maybe I'll consolidate these.

rrtypes.go
```diff
@@ -0,0 +1,130 @@
...
+// to provide clients with information for establishing HTTPS
+// connections to servers. It may include data about ALPN,
+// ECH, IP hints, and more.
+type HTTPS struct {
+	Name     string
+	TTL      time.Duration
+	Priority uint16
+	Target   string
+	Value    SvcParams
+}
```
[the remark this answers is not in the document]
mholt: Is Scheme unique to the SVCB type? Or are you thinking of a 3-in-1 struct here (SVCB, HTTPS, and SRV)?

rrtypes.go
```diff
@@ -0,0 +1,130 @@
...
+
+// TXT represents a parsed TXT-type record, which is used to
+// add arbitrary text data to a name in a DNS zone. It is often
+// used for email integrity (DKIM/SPF), site verification, ACME
+// challenges, and more.
+type TXT struct {
+	Name string
+	TTL  time.Duration
+	Text string
+}
```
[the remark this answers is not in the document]
mholt: I think so; because if you're type-switching you would switch on the type... granted, I don't have types (yet) for all RR-types (and probably never will), TXT is a pretty common one, so I think it's probably worth having a type for? I dunno. You might be right, but when I thought of the type-switching scenarios, I felt like it was a good idea to do this.

rrtypes.go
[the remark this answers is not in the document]
gucci-on-fleek: > Yeah, good idea... I guess. Feels a little weird to break the pattern though and have a RR type that isn't named the same.

Yeah, it's definitely a little weird, but I think that it would be a nicer API, probably?

> Are there any other semantic differences between the two record types that you know of?

Nope. There [used to be `A6` records](https://datatracker.ietf.org/doc/html/rfc6563) which had the same role as `A` records but with special v6 semantics, but those are obsolete (and were never really used anyways). All that's used these days are `AAAA` records, and [those behave identically to `A` records](https://datatracker.ietf.org/doc/html/rfc3596) (except with a v6 address instead of v4),

rrtypes.go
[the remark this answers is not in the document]
gucci-on-fleek: > Is Scheme unique to the SVCB type? 

Yeah, `Scheme` is unique to `SVCB` (and `HTTPS`). An [`SVCB` scheme](https://datatracker.ietf.org/doc/html/rfc9460#section-2.3) is essentially the same thing as an [`SRV` service](https://datatracker.ietf.org/doc/html/rfc6335#section-5.2), but they use different protocol registries ([`SVCB`](https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#underscored-globally-scoped-dns-node-names), [`SRV`](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml)). 

I'm not sure which would be the least confusing option here though. One option is to use different names for each: `Scheme` for `SVCB` and `Service` for `SRV`. This matches the RFCs, so these are the names that people are familiar with. But this also might be confusing since we'd have two different names for what are effectively the same thing. The other option would be to pick one of `Scheme`/`Service` and use it for both RR types. This would make it obvious that both fields refer to the same thing and would give a more consistent API. But this might be confusing since that means that `libdns`'s names won't match up with the names used elsewhere.

> Or are you thinking of a 3-in-1 struct here (SVCB, HTTPS, and SRV)?

I think that `SRV` and `SVCB` need to be kept separate; otherwise, someone might try to add a `Weight` to a `SVCB` or a `ech=` to a `SRV`, neither of which are valid.

rrtypes.go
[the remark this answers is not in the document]
gucci-on-fleek: I think that only having specific structs for the RR types with special fields makes sense, and I think that having specific structs for all (common) RR types also makes sense, but I think that choosing somewhere in between might be confusing.

If we want to have Go-types for all the most common RR-types, I think that we'd only need to add `CAA`, `MX`, and `NS` RR-types, since I don't think that there are any other common ones. Or, you could only have `Address`, `SVCB`, and `SRV` Go-types since those are all special and common. Both options seem reasonable to me.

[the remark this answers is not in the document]
mholt: I agree -- my goal is to have struct types for all _common_ RR types. I don't have them all implemented here since it's still experimental, but we can definitely add more later (or anyone can contribute more). I personally do not use CAA, MX, or NS types myself, so maybe it'd be better if someone else contributed those.

rrtypes.go
```diff
@@ -0,0 +1,130 @@
...
+//
+// EXPERIMENTAL; subject to change or removal.
+func (h HTTPS) RR() (RR, error) {
+	return RR{
+		Name: h.Name,
+		TTL:  h.TTL,
+		Type: "HTTPS",
+		Data: fmt.Sprintf("%d %s %s", h.Priority, h.Target, h.Value),
+	}, nil
+}
```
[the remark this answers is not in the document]
mholt: Oh, I see -- this kind of dovetails with the above comments.

I think I get what you mean now; but I also think that due to the differences in the semantics around Name, maybe HTTPS and SVCB should remain separate types?

mholt:
@gucci-on-fleek Thanks for the thorough review!

I believe I have addressed (pun intended?) all the comments, including combining A and AAAA into an Address struct, except for the matter of SVCB records.

Real quick: the two downsides to combining into the Address type, that I can think of, are:

- Users might be looking for A or AAAA in the godoc, and won't find it, unless they do a Find-in-Page. (Simply pressing "F" won't bring up `Address`, at least I assume not, because "A" and "AAAA" are not in the type definition itself.)
- There is no validation to ensure IPv4 or IPv6.

However, given that users don't really care what the size of the address is in Go code, I think at least that second downside is acceptable. The code automatically assigns the correct RR-type either way.

---

As for SVCB records, I'm inclined to keep them separate since they have slightly different semantics around the name, compared to HTTPS records. Do you agree?

rrtypes.go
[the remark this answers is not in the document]
gucci-on-fleek: > I'm leaning towards Scheme/Service, since they are going to be in different struct types anyway, so we might as well use the terminology people are accustomed to specifically for that type.

:+1: 

> However, is this naming pattern typical of HTTPS records too? 

An `HTTPS` record is just a `SVCB` record with an implicit `https` scheme (see [RFC 9460 §B](https://datatracker.ietf.org/doc/html/rfc9460#appendix-B)). So both of the following records should be semantically equivalent:

```zone
www         3600  IN  HTTPS  1  .  alpn=h2,h3
_https.www  3600  IN  HTTPS  1  .  alpn=h2,h3
```

except that (see [RFC 9460 §9](https://datatracker.ietf.org/doc/html/rfc9460#section-9)) specifically forbids using `SVCB` with a scheme of `http` or `https`.

> If not, the different semantics around the Name make me want to keep SVCB and HTTPS separate.

I think that this is fairly similar to the `A`/`AAAA` case: both `SVCB` and `HTTPS` have the same semantics in every way except for the scheme, so I think that it would be better to combine them. But that being said, this is a bit of a weird case, so I think that it would be entirely reasonable to keep them separate as well.

rrtypes.go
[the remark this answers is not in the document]
gucci-on-fleek: > Oh, I see -- this kind of dovetails with the above comments.
> 
> I think I get what you mean now; but I also think that due to the differences in the semantics around Name, maybe HTTPS and SVCB should remain separate types?

Well, RFC 9460 explicitly provides for new RR types with `SVCB` semantics to be created (like `HTTPS`). So in theory, one day there might be 5 different RR types like this, which would be a pretty strong motivation to combine them all so that there are fewer special cases. But in practice, I can't imagine that anyone will ever create another `SVCB`-like RR type, and probably 95% of `libdns` users will only ever use `HTTPS` records and not `SVCB` records. So I'd personally lean towards one combined `SVCB`/`HTTPS` Go type, but I think that having two separate `SVCB`/`HTTPS` Go types would be entirely reasonable too.

rrtypes.go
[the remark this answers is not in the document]
gucci-on-fleek: > my goal is to have struct types for all _common_ RR types. I don't have them all implemented here since it's still experimental

:+1:

rrtypes.go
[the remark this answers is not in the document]
mholt: > www         3600  IN  HTTPS  1  .  alpn=h2,h3
> _https.www  3600  IN  HTTPS  1  .  alpn=h2,h3

(Did you mean the second line to be "SVCB" instead of "HTTPS"?

Anyway... yeah... that's very interesting though, and... frustrating at the same time.

Would you be interested in proposing a PR that makes SVCB and HTTPS records the same type? I might like to see what that looks like. I was surprised that I accepted the `Address` type suggestion, so maybe it's something similar here.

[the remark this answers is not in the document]
gucci-on-fleek: > > www         3600  IN  HTTPS  1  .  alpn=h2,h3
> > _https.www  3600  IN  HTTPS  1  .  alpn=h2,h3
> 
> (Did you mean the second line to be "SVCB" instead of "HTTPS"?

Haha, yes, sorry about that.

> Would you be interested in proposing a PR that makes SVCB and HTTPS records the same type? I might like to see what that looks like. I was surprised that I accepted the `Address` type suggestion, so maybe it's something similar here.

Sure.

mholt:
I think we're close. Just gotta figure out the SVCB/HTTPS record types and then I'd be good to merge it. Thanks for collaborating with me on it!

libdns.go
```diff
@@ -404,13 +222,13 @@ func RelativeName(fqdn, zone string) string {
...
-// prepending it to “zone” and tidying up the dots. For example, an input of
+// [AbsoluteName] makes name into a fully-qualified domain name (FQDN) by
+// prepending it to zone and tidying up the dots. For example, an input of
 // name “sub” and zone “example.com.” will return “sub.example.com.”.
 //
 // Using “@” as the name is the recommended way to represent the root of the
 // zone; however, unlike the [Record] struct, using the empty string "" for the
-// name *is* permitted here, and will be identically to “@”.
+// name *is* permitted here, and will be treated identically to “@”.
 func AbsoluteName(name, zone string) string {
```
gucci-on-fleek, line 232: While porting `libdns/rfc2136` to use the new interfaces in this PR, one problem that I noticed is that calling `AbsoluteName` on an already-absolute domain (with the trailing dot) still appends the zone to the domain name. I think that this should behave like [`path/filepath.Abs`](https://pkg.go.dev/path/filepath#Abs) and do nothing if the domain name is already absolute.

```suggestion
func AbsoluteName(name, zone string) string {
	if zone == "" {
		return strings.Trim(name, ".")
	}
	if name == "" || name == "@" {
		return zone
	}
	if strings.HasSuffix(name, ".") {
		// Already a FQDN, so just return it
		return name
	}
	return name + "." + zone
}
```

gucci-on-fleek (commented):
Ok, [I ported `libdns/rfc2136` to use the new interfaces in this PR](https://github.com/libdns/rfc2136/compare/master...gucci-on-fleek:libdns-rfc2136:libdns-1.0), and everything went pretty well, which I think is a good sign that the new semantics/interfaces are beneficial.

The only challenge that I had was converting all `Target` fields to FQDNs in the test code. This was a little bit awkward since I had to hardcode the list of types with target fields:

https://github.com/gucci-on-fleek/libdns-rfc2136/blob/4d49bd/provider_test.go#L66-L89

But testing code is often pretty awkward, so I think that this is fine. And similarly, [`==` doesn't work on the `HTTPS` structs](https://github.com/gucci-on-fleek/libdns-rfc2136/blob/4d49bd3682f7ddf0b6688bc4c5b939222f7bcc51/provider_test.go#L91-L108), but I don't think that you ever need to compare records outside of testing code, so I don't think that there's any need to add an `Equal` method.

> Would you be interested in proposing a PR that makes SVCB and HTTPS records the same type? I might like to see what that looks like.

> Just gotta figure out the SVCB/HTTPS record types and then I'd be good to merge it

I've got an exam tomorrow, so I'll be busy studying tonight, but I've got time tomorrow evening, so I'll submit a PR then.

Thanks!

record.go
```diff
@@ -0,0 +1,444 @@
...
+	"fmt"
+	"net/netip"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// Record is any type that can reduce itself to the [RR] struct.
+type Record interface {
+	RR() (RR, error)
```
mjl-, line 13: Couldn't this be `RR() RR`. What could the error be? I've done a quick attempt to change dnsclay to use this new interface, and I have to do `rr, _ := record.RR()` everywhere. The Record-implementing types all have an `RR()` that always returns success. The RR.Parse function seems to be for parsing in a way that can return errors.

record.go
```diff
@@ -0,0 +1,444 @@
...
+	// especially: A, AAAA, CNAME, TXT, HTTPS, and SRV.
+	//
+	// Other custom record types may be supported with implementation-defined
+	// behavior.
+	Type string
+
+	// The data (or "value") of the record. This field should be formatted in the
+	// *unescaped* standard zone file syntax (technically, the "RDATA" field as
+	// defined by RFC 1035 §5.1). Due to variances in escape sequences and provider
+	// support, this field should not contain escapes.
```
mjl-, line 69: Just wondering: What kind of escapes are meant here?
For example https://datatracker.ietf.org/doc/html/rfc3597 has RDATA values with what I would think are escape sequences, e.g. `e.example.   IN          A               \# 4 0A000001`

libdns.go
```diff
@@ -1,394 +1,212 @@
...
+	// SetRecords updates the zone so that the records described in the input are
+	// reflected in the output. It may create or update records or—depending on
+	// the record type—delete records to maintain parity with the input. No other
+	// records are affected. It returns the records which were set.
+	//
+	// For any (name, type) pair in the input, SetRecords ensures that the only
+	// records in the output zone with that (name, type) pair are those that were
+	// provided in the input.
+	//
+	// In RFC 7719 terms, SetRecords appends, modifies, or deletes records in the
```
mjl-, line 101: Not really related, but this could be a good moment to update the reference to  RFC 9499. The updates have been: 7719 -> 8499 -> 9499. See https://datatracker.ietf.org/doc/html/rfc9499 and follow the "Obsoletes" references in the pane on the right. The RFC is referenced earlier in this file too, and perhaps other places.

libdns.go
```diff
@@ -1,394 +1,212 @@
...
+// [libdns.RecordAppender.AppendRecords] is called simultaneously, and two API
+// requests are made to the provider at the same time, the result of both requests
+// must be visible after they both complete; if the provider does not synchronize
+// the writing of the zone file and one request overwrites the other, then the
+// client implementation must take care to synchronize on behalf of the incompetent
+// provider. This synchronization need not be global; for example: the scope of
+// synchronization might only need to be within the same zone, allowing multiple
+// requests at once as long as all of them are for different zone. (Exact logic
+// depends on the provider.)
+//
```
mjl-, line 50: Can we mention providers can optionally implement io.Closer, for when they keep state around that should be cleaned up when done with the provider. Or perhaps a `Close()` method without error? Useful with providers that create an http.Client for which idle connections are kept around. Their `Close` would call `http.Client.CloseIdleConnections`.

Having

libdns.go
```diff
@@ -1,394 +1,212 @@
...
-// “AppendRecords” creates the requested records in the given zone and returns
-// the populated records that were created. It never changes existing records.
-// Therefore, it is invalid to use this method with “CNAME”-type records.
-//
-// Implementations must honor context cancellation and be safe for concurrent
-// use.
 type RecordAppender interface {
+	// AppendRecords creates the inputted records in the given zone and returns
+	// the populated records that were created. It never changes existing records.
+	// Therefore, it is invalid to use this method with CNAME-type records.
```
mjl-, line 83: I know this was added in a previous PR, but with a revamp and same people involved in this PR: Is the intend to say that you can't add a CNAME record when you already have other types of records with the same name? Provider implementers could interpret this line to say they should just reject any attempt to add a CNAME record with AppendRecords. I would think we can just "append" a single CNAME record for a name that doesn't exist yet.

libdns.go
```diff
@@ -1,394 +1,212 @@
...
-//	a.example.com. 3600 IN AAAA 2001:db8::1
-//	a.example.com. 3600 IN AAAA 2001:db8::2
-//	a.example.com. 3600 IN AAAA 2001:db8::5
-//	b.example.com. 3600 IN AAAA 2001:db8::3
-//	b.example.com. 3600 IN AAAA 2001:db8::4
 type RecordSetter interface {
+	// SetRecords updates the zone so that the records described in the input are
+	// reflected in the output. It may create or update records or—depending on
+	// the record type—delete records to maintain parity with the input. No other
+	// records are affected. It returns the records which were set.
```
mjl-, line 95: Should users/implementers of this function consider the CNAME case? What should if I "set" a CNAME record when the name already exists with another type? This isn't valid in DNS (except for some special records I believe, like NS and DNSKEY at the apex). Options:
1. Ignore the request for setting the new CNAME record (I think this is the behaviour in DNS UPDATE)
2. Replace the existing records and create the CNAME (this isn't allowed by the current wording)
3. Require that an error be returned.
4. Provider-specific behaviour? Where some providers cause an error, others could ignore the request, yet others could replace other existing records with the same name with a CNAME (this is not allowed by current wording).

libdns.go
```diff
@@ -1,394 +1,212 @@
...
-//	a.example.com. 3600 IN AAAA 2001:db8::5
-//
-//	;; Resultant zone
-//	a.example.com. 3600 IN AAAA 2001:db8::1
-//	a.example.com. 3600 IN AAAA 2001:db8::2
-//	a.example.com. 3600 IN AAAA 2001:db8::5
-//	b.example.com. 3600 IN AAAA 2001:db8::3
-//	b.example.com. 3600 IN AAAA 2001:db8::4
 type RecordSetter interface {
+	// SetRecords updates the zone so that the records described in the input are
```
mjl-, line 92: I suppose it's out of the question to rename this method to SetRecordSets? It should make it clear that the behaviour is quite a bit different/more special than the Append and Delete methods. I'm afraid users/implementers of this method may make the wrong assumption based on the name...

mjl- (commented):
Hi Matt & gucci-on-fleek, thanks for the awesome work! The Record/RR approach looks good.
I looked at the changes I need to make in dnsclay to work with this update, and it's very limited.
I added a few comments/questions in the PR, hope they help.

mjl-:
What about adding some test helpers to libdns/libdns, or subpackage, or other testing-module, that provider implementers can run on their provider? I think it would be like what is suggested in issue #44. I can write some of these tests.

I imagine the tests would be run with a provider impelmentation, zone name, and "subzone" name, and the tests would get/add/set/delete records and see the changes are reflected properly, and check the names are well formed (relative names, "@" instead of empty name, nothing outside the zone).

We could add an Options struct to influence behaviour to check for, such as skipping checks for whether exact TTLs are set (some DNS providers appear to only allow setting a small set of TTLs). The options can be expanded later on.

We could check context cancellations, do concurrent requests and test with the race checker on. We would take care not to touch anything outside the subzone, but of course bugs in provider implementations could do the wrong thing... Hopefully implementers are in a position to create an isolated account/zone.

libdns.go
mholt: So that fails a test case, where the name is `"www."` and the zone is `"example.com."` -- should the output really be `www.`? Should the name be considered absolute only if the zone is empty string (and maybe also ".")? Just trying to figure out what makes the most sense.

record.go
mholt: I did at first, but then I was like "You know, not returning an error got me in trouble with a couple exported APIs when I designed Caddy, so it doesn't hurt to do so... I might as well here." That was before any RR struct types had been implemented.

Now that several RR types have been implemented, I think it's more obvious to me that they should always be able to reduce to an RR, without error. (That is, after all, the definition of a type that implements the Record interface. If it has an error, it's like... not even a DNS record??)

So yeah, maybe we can get rid of the error return value here. What do you think @gucci-on-fleek ?

record.go
mholt: Hmm, well I think since escape sequences can be "parsed" or "evaluated" to become their _actual_ values, we intend that this field always contains the parsed/evaluated contents. So, maybe _any_ escapes? I dunno though... @gucci-on-fleek might have good thoughts on this.

libdns.go
mholt: Good point... yeah I'd be okay with updating it.

record.go
mholt: I went ahead and pushed a commit that makes it `RR() RR`, no error.

I _assume_ this will be okay, although now I remember one instance where an error WAS returned... when `type Address` used to be two separate types, `A` and `AAAA`, in an earlier revision, if the `netip.Addr` field wasn't the respective IPv4 or IPv6 bit length, then there would have been an error.

Without returning an error, I'm not sure what we'd have done. Just let it happen and expect the DNS provider API to return an error when trying to put an IPv6 address in an A record or something?

So, while it seems to be okay _currently_, I am super curious if there's ever a possibility that an error would be useful. If so, I'd like to restore it.

libdns.go
mholt: I think the intent is that provider packages don't have to worry about this. Basically, using AppendRecords() is error-prone when appending CNAME records.

Although, as you state below, using SetRecords with CNAME is also error-prone, since another record type could already exist for the domain name.

What if we just revise the comment here to advise caution when appending CNAMEs? (Ultimately, the DNS service API should return an error, I would think, but we don't know that for sure. So it might be best if the user checks to ensure there isn't another record for the domain name already, before calling Append or Set for a CNAME record.)

libdns.go
mholt: Hm, yeah I dunno, I like the current naming convention because it's consistent.

I also like the "play" on words: Set refers to both the noun and verb form. (Makes a lot of sense when you think about it, set applies the definition of a set.)

libdns.go
gucci-on-fleek: > where the name is `"www."` and the zone is `"example.com."` -- should the output really be `www.`? 

Yes, the trailing dot means that this is already a FQDN, so `www.` is a valid absolute name. There isn't any domain in the root zone called `www.` right now, but it would ostensibly be a valid name. This isn't strictly theoretical—there are actually a few TLDs with `A` records set. Up until very recently [there used to be quite a few of these](https://en.wikipedia.org/wiki/Top-level_domain#Dotless_domains), but `uz` still works:

- [`https://uz./`](https://uz./)
- `curl --insecure -vvv --doh-url 'https://dns.google/dns-query' https://uz./`

> Should the name be considered absolute only if the zone is empty string (and maybe also ".")? Just trying to figure out what makes the most sense.

The main advantage to doing nothing if there's a trailing dot is that `AbsoluteName` becomes idempotent. So with the following example

```go
package main

import (
	"fmt"

	"github.com/libdns/libdns"
)

func main() {
	zone := "example.com."
	rrs := []libdns.CNAME{
		{
			Name:   "one",
			TTL:    3600,
			Target: "two",
		},
		{
			Name:   "three",
			TTL:    3600,
			Target: "four.example.net.",
		},
	}

	for _, rr := range rrs {
		fmt.Println(libdns.AbsoluteName(rr.Target, zone))
	}
}
```

with this change you'd get

```txt
two.example.com.
four.example.net.
```

but _without_ this change, you'd need the following code to get the same output:

```go
package main

import (
	"fmt"
	"strings"

	"github.com/libdns/libdns"
)

func main() {
	zone := "example.com."
	rrs := []libdns.CNAME{
		{
			Name:   "one",
			TTL:    3600,
			Target: "two",
		},
		{
			Name:   "three",
			TTL:    3600,
			Target: "four.example.net.",
		},
	}

	for _, rr := range rrs {
		target := rr.Target
		if !strings.HasSuffix(target, ".") {
			target = libdns.AbsoluteName(rr.Target, zone)
		}
		fmt.Println(target)
	}
}
```

This change would also make `libdns.AbsoluteName` behave similarly to `path/filepath.Abs`, where you can call it on a path that begins with `/` and it will do nothing:

```go
package main

import (
	"fmt"
	"os"
	"path/filepath"
)

func main() {
	os.Chdir("/tmp/")
	paths := []string{
		"folder/file.txt",
		"/etc/passwd",
	}
	for _, path := range paths {
		path, _ = filepath.Abs(path)
		fmt.Println(path)
		// Call it 4 more times
		for i := 0; i < 4; i++ {
			path, _ = filepath.Abs(path)
		}
		fmt.Println(path)
	}
}
```

```txt
/tmp/folder/file.txt
/tmp/folder/file.txt
/etc/passwd
/etc/passwd
```

record.go
gucci-on-fleek: > Couldn't this be `RR() RR`. What could the error be? I've done a quick attempt to change dnsclay to use this new interface, and I have to do `rr, _ := record.RR()` everywhere. 

Yeah, I also had to do `rr, _ := record.RR()`, so this is a :+1: from me.

libdns.go
gucci-on-fleek: Oh wow, I don't know how I missed that chain of “obsoleted by”s; good catch. Plus, RFC 9499 is way easier to understand, so I'll definitely update this.

libdns.go
gucci-on-fleek: Somewhat related to the batching discussions in #154.

libdns.go
gucci-on-fleek: > I would think we can just "append" a single CNAME record for a name that doesn't exist yet.

Err, yes. I'm not sure how I missed that rather obvious case. 

> Is the intend to say that you can't add a CNAME record when you already have other types of records with the same name?  

Since it's not always obvious when to use `AppendRecords` and when to use `SetRecords`, my intent was to try and provide some concrete guidelines. 

> What if we just revise the comment here to advise caution when appending CNAMEs?

Sure, I'll add that in the next commit.

libdns.go
gucci-on-fleek: > Should users/implementers of this function consider the CNAME case?

Good catch; yes, they probably should.

> 4. Provider-specific behaviour?

Yeah, I think that that's the only way to go here, since the other options could require multiple `(Get/Delete)Records` call internally.

> 2. Replace the existing records and create the CNAME

Seems like a reasonable choice for an implementation.

> 3. Require that an error be returned.

Also seems like a reasonable choice.

> 1. Ignore the request for setting the new CNAME record (I think this is the behaviour in DNS UPDATE)

Hmm, even though [I'm working on porting this PR to the `libdns` DNS UPDATE provider](https://github.com/libdns/rfc2136/compare/master...gucci-on-fleek:libdns-rfc2136:libdns-1.0) and allowing this would make things easier for me, I feel like we should disallow this specific case, because it feels like a really bad idea for us to allow a call to `SetRecords` to do nothing yet still throw no errors.

---

I'll add a paragraph about this CNAME stuff in my next commit, so please let me know what you think.

libdns.go
gucci-on-fleek: > I suppose it's out of the question to rename this method to SetRecordSets?

I think that if you know enough about DNS to know what an RRset is, you would know to be suspicious of methods named `SetRecords` and fully read their documentation before using them. `SetRecordSets` would definitely be a more correct name, but I still don't think that that name makes it obvious what the method does without first reading its documentation.

rrtypes.go
```diff
@@ -0,0 +1,190 @@
...
+		Data: fmt.Sprintf("%d %s %s", h.Priority, h.Target, h.Value),
+	}
+}
+
+// MX represents a parsed MX-type record, which is used to specify the hostnames
+// of the servers that accept mail for a domain.
+type MX struct {
+	Name       string
+	TTL        time.Duration
+	Preference uint16 // Lower values indicate that clients should prefer this server
```
gucci-on-fleek, line 106: What should this field be called? The RFCs call it `Preference`, but since `SRV` and `HTTPS` use `Priority`, I'm thinking that maybe we should call it `Priority` here for consistency?

gucci-on-fleek:
Ok, I've pushed support for `SVCB` records. The code is a bit of a mess, so please feel free to change it as you wish. All of the current tests pass, but I'll add a few more tests tomorrow. Anyways, please let me know what you think.

rrtypes.go
mholt: Personally I'd prefer `Preference` since that's what the RFC for MX calls it.

The name doesn't have to be the same as in the other types, since this is a concrete type that requires its own code anyways, so it's not like an interface where it has to be the same for all types that implement it. :+1:

rrtypes.go
```diff
@@ -0,0 +1,374 @@
...
+	// instead treat the entire TXT as a single, arbitrary-length string. libdns
+	// implementations are therefore expected to handle this as required by
+	// their respective DNS provider APIs. See the [DNSControl explainer] on
+	// this for more information.
+	//
+	// [RFC 1035 §3.3.14]: https://datatracker.ietf.org/doc/html/rfc1035#section-3.3.14
+	// [RFC 7208 §3.3]: https://datatracker.ietf.org/doc/html/rfc7208#section-3.3
+	// [DNSControl explainer]: https://docs.dnscontrol.org/developer-info/opinions#opinion-8-txt-records-are-one-long-string
+	Text string
+}
```
gucci-on-fleek, line 365: While porting over `libdns/rfc2136`, I noticed that [it tests for an obscure edge case for `TXT` records](https://github.com/libdns/rfc2136/blob/fc4796f45d00437685869ef53d83125b22465a4e/conversion_test.go#L20-L48). This seems like more of an RFC-trivia thing than something that users will actually have a need for, so I think that it makes the most sense to just forbid this. 

I guess the other option would be to instead have

```go
type TXT struct {
	Name string
	TTL  time.Duration
	Text []string
}
```

but then we'd have to explain to users that `one` and `two` are completely different:

```go
one := []libdns.TXT{
	{
		Name: "@",
		TTL:  time.Hour,
		Text: []string{
			"v=spf1 include:_spf.example.com ~all",
			"google-site-verification=abc123",
		},
	},
}
two := []libdns.TXT{
	{
		Name: "@",
		TTL:  time.Hour,
		Text: []string{"v=spf1 include:_spf.example.com ~all"},
	},
	{
		Name: "@",
		TTL:  time.Hour,
		Text: []string{"google-site-verification=abc123"},
	},
}
```

libdns.go
```diff
@@ -1,394 +1,224 @@
...
-// “MX”, or “CNAME” records, which store their targets in the
-// [libdns.Record.Value] field. This field must be set to the fully-qualified
-// domain name (FQDN) of the target (which includes the trailing dot).
-//
-// Examples:
-//   - (HTTPS) “example.com.”
-//   - (SVCB)  “example.com.”
-type Record struct {
-	// provider-specific metadata
-	ID string
```
devhaozi, line 303: Q: If the ID is removed, most DNS providers will have to query all records when deleting and updating records, and then iterate through the record type, record value, etc.

gucci-on-fleek: > If the ID is removed

Providers are still allowed to include IDs (and any other fields that they want) in their `Record` types

https://github.com/libdns/libdns/blob/1e63e027545500a1dabe8f256b74a8c838ef26af/record.go#L11-L14

but now all that's mandatory is that you can convert this to an `RR`, which only contains `Name`, `TTL`, `Type`, and `Data`:

https://github.com/libdns/libdns/blob/1e63e027545500a1dabe8f256b74a8c838ef26af/record.go#L16-L90

> most DNS providers will have to query all records when deleting and updating records, and then iterate through the record type, record value, etc to get record ID

Right, but the problem with IDs is that not all providers assign a unique ID to each record, and it's not always possible for a `libdns` implementation to synthesize an ID. See https://github.com/libdns/libdns/pull/152#discussion_r1990509576 for an example of where this can go wrong. 

Another issue with IDs is that not every provider supports the same operations on a record with a constant ID. Assuming that a provider assigns unique IDs to all records, then changing its `Data` is always going to be supported, but what happens if we change the `Name` or `Type`? In that case, it's better to keep things explicit and force the `libdns` API user to use `(Append/Set/Delete)Records` as necessary. 

> In previous versions, most use cases of libdns were to create a new record -> return the record ID -> wait for DNS validation to complete -> delete the record by ID.

Yes, you can no longer delete a record by its ID, but you can still either delete the records returned by `SetRecords`, or you can set the `Record` fields to the special “wildcard” values:

https://github.com/libdns/libdns/blob/1e63e027545500a1dabe8f256b74a8c838ef26af/libdns.go#L182-L191

Example:

```go
package main

import (
	"context"

	"github.com/libdns/libdns"
	"github.com/libdns/rfc2136"
)

func GetCertificate(domain string, zone string) ([]byte, error) {
	ctx := context.Background()
	provider := rfc2136.Provider{}
	acmeValue := acme.GetChallengeValue()
	domain = "_acme-challenge." + domain

	recs, err := provider.SetRecords(ctx, zone, []libdns.Record{libdns.TXT{
		Name: domain,
		TTL:  0,
		Text: acmeValue,
	}})
	if err != nil {
		return nil, err
	}

	certificate, err := acme.GetCertificate()
	if err != nil {
		return nil, err
	}

	// Option 1
	_, err = provider.DeleteRecords(ctx, zone, recs)

	// // Option 2
	// _, err = provider.DeleteRecords(ctx, zone, []libdns.Record{libdns.TXT{
	//     Name: domain,
	// }})
	if err != nil {
		return nil, err
	}

	return certificate, nil
}
```

---

That being said, nothing in v1.0 has been finalized yet, so if you _really_ think that no ID field is going to be a problem, then we can reconsider. But since `libdns` aims to support the “lowest common denominator” of all providers, we'd need to figure out some way to make the non-ID-supporting providers behave the same as the ID-supporting providers, and I'm not sure if that's possible.

libdns.go
```diff
@@ -1,394 +1,224 @@
...
-	Target   string
+// [ZoneGetter] can list available DNS zones.
+type ZoneGetter interface {
+	// GetZones returns the list of available DNS zones for use by other [libdns]
+	// methods.
+	//
+	// Implementations must honor context cancellation and be safe for concurrent
+	// use.
+	GetZones(ctx context.Context) ([]Zone, error)
 }
```
gucci-on-fleek, line 210: 1. Why was this renamed from `ListZones` to `GetZones`? `ListZones` feels like a slightly better name to me since it's more obvious what it does, but there is a nice symmetry between `GetZones` and `GetRecords`.

2. I'm thinking that we might want to drop `ListZones`/`GetZones` for v1.0. [Only 3 `libdns` providers currently implement `ListZones`](https://github.com/search?q=%28org%3Alibdns+OR+org%3Acaddy-dns%29+AND+%28NOT+path%3A%2Fexample%7Ctest%2F%29+ListZones&type=code), and across all of GitHub, [only 2 repositories currently use `ListZones`](https://github.com/search?q=NOT+org%3Alibdns+NOT+is%3Afork+%22github.com%2Flibdns%22+%22ListZones%22&type=code). Both of these repositories hardcode in a single `libdns` implementation ([1](https://github.com/takingnames/namedrop-go/blob/26a0fc69b8dfc31d9e5c14a3d3af3aa1f4eb7480/cmd/namedrop-server/main.go#L254-L265), [2](https://github.com/waygate-io/waygate-go/blob/9b5b4fecb7f7b5561dea68ae354bf611c508a9e6/client.go#L500-L507)), so removing support from `libdns` should have fairly little impact.

    That being said, it's not like `ListZones`/`GetZones` takes much code or has any maintenance impact, so it would be harmless to keep it around, even if it's not being used.

libdns.go
devhaozi: > First, thanks for the feedback; it's good to hear from provider implementers so that we know if we're heading in the right direction or not.
> 
> > If the ID is removed
> 
> Providers are still allowed to include IDs (and any other fields that they want) in their `Record` types
> 
> https://github.com/libdns/libdns/blob/1e63e027545500a1dabe8f256b74a8c838ef26af/record.go#L11-L14
> 
> but now all that's mandatory is that you can convert this to an `RR`, which only contains `Name`, `TTL`, `Type`, and `Data`:
> 
> https://github.com/libdns/libdns/blob/1e63e027545500a1dabe8f256b74a8c838ef26af/record.go#L16-L90
> 
> > most DNS providers will have to query all records when deleting and updating records, and then iterate through the record type, record value, etc to get record ID
> 
> Right, but the problem with IDs is that not all providers assign a unique ID to each record, and it's not always possible for a `libdns` implementation to synthesize an ID. See [#152 (comment)](https://github.com/libdns/libdns/pull/152#discussion_r1990509576) for an example of where this can go wrong.
> 
> Another issue with IDs is that not every provider supports the same operations on a record with a constant ID. Assuming that a provider assigns unique IDs to all records, then changing its `Data` is always going to be supported, but what happens if we change the `Name` or `Type`? In that case, it's better to keep things explicit and force the `libdns` API user to use `(Append/Set/Delete)Records` as necessary.
> 
> > In previous versions, most use cases of libdns were to create a new record -> return the record ID -> wait for DNS validation to complete -> delete the record by ID.
> 
> Yes, you can no longer delete a record by its ID, but you can still either delete the records returned by `SetRecords`, or you can set the `Record` fields to the special “wildcard” values:
> 
> https://github.com/libdns/libdns/blob/1e63e027545500a1dabe8f256b74a8c838ef26af/libdns.go#L182-L191
> 
> Example:
> 
> ```go
> package main
> 
> import (
> 	"context"
> 
> 	"github.com/libdns/libdns"
> 	"github.com/libdns/rfc2136"
> )
> 
> func GetCertificate(domain string, zone string) ([]byte, error) {
> 	ctx := context.Background()
> 	provider := rfc2136.Provider{}
> 	acmeValue := acme.GetChallengeValue()
> 	domain = "_acme-challenge." + domain
> 
> 	recs, err := provider.SetRecords(ctx, zone, []libdns.Record{libdns.TXT{
> 		Name: domain,
> 		TTL:  0,
> 		Text: acmeValue,
> 	}})
> 	if err != nil {
> 		return nil, err
> 	}
> 
> 	certificate, err := acme.GetCertificate()
> 	if err != nil {
> 		return nil, err
> 	}
> 
> 	// Option 1
> 	_, err = provider.DeleteRecords(ctx, zone, recs)
> 
> 	// // Option 2
> 	// _, err = provider.DeleteRecords(ctx, zone, []libdns.Record{libdns.TXT{
> 	//     Name: domain,
> 	// }})
> 	if err != nil {
> 		return nil, err
> 	}
> 
> 	return certificate, nil
> }
> ```
> 
> That being said, nothing in v1.0 has been finalized yet, so if you _really_ think that no ID field is going to be a problem, then we can reconsider. But since `libdns` aims to support the “lowest common denominator” of all providers, we'd need to figure out some way to make the non-ID-supporting providers behave the same as the ID-supporting providers, and I'm not sure if that's possible.

Thanks for clarifying this, IDs are important among DNS providers in China because they support subnet providers and subregional resolution. For example, without IDs, there is no way to delete a single record for China Telecom/Beijing/Shanghai. (Although this is an edge case, it is frequently used in Chinese DNS providers)

gucci-on-fleek: > Thanks for clarifying this, IDs are important among DNS providers in China because they support net provider-based and region-based resolution. For example, without IDs, there is no way to delete a single record for China Telecom/Beijing/Shanghai. (Although this is an edge case, it is frequently used in Chinese DNS providers)

How does that work with `AppendRecords` if you're adding a brand-new record? Because if I'm understanding you correctly, you'd need to set the location for that record, but since a `libdns.Address` only has `(Name, TTL, IP)` fields, there's no obvious way to do that.

libdns.go
mholt: > Why was this renamed from ListZones to GetZones? ListZones feels like a slightly better name to me since it's more obvious what it does, but there is a nice symmetry between GetZones and GetRecords.

Yeah, just for symmetry with GetRecords. Although, in hindsight, maybe GetRecords should have been called ListRecords, since there's no parameter that filters what you're getting, which I feel like is typical of "getters".

So... I dunno. It's purely an aesthetic change for symmetry.

> I'm thinking that we might want to drop ListZones/GetZones for v1.0. [Only 3 libdns providers currently implement ListZones](https://github.com/search?q=%28org%3Alibdns+OR+org%3Acaddy-dns%29+AND+%28NOT+path%3A%2Fexample%7Ctest%2F%29+ListZones&type=code), and across all of GitHub, [only 2 repositories currently use ListZones](https://github.com/search?q=NOT+org%3Alibdns+NOT+is%3Afork+%22github.com%2Flibdns%22+%22ListZones%22&type=code). Both of these repositories hardcode in a single libdns implementation ([1](https://github.com/takingnames/namedrop-go/blob/26a0fc69b8dfc31d9e5c14a3d3af3aa1f4eb7480/cmd/namedrop-server/main.go#L254-L265), [2](https://github.com/waygate-io/waygate-go/blob/9b5b4fecb7f7b5561dea68ae354bf611c508a9e6/client.go#L500-L507)), so removing support from libdns should have fairly little impact.

Thanks for the analysis! I figured it was only used a little bit, but I didn't realize that even a second repo used it.

While all the libdns interfaces are technically optional, we could mark this one explicitly as optional, since I agree, its utility is quite limited and many providers may not have a way of listing zones anyway.

libdns.go
devhaozi: > > Thanks for clarifying this, IDs are important among DNS providers in China because they support net provider-based and region-based resolution. For example, without IDs, there is no way to delete a single record for China Telecom/Beijing/Shanghai. (Although this is an edge case, it is frequently used in Chinese DNS providers)
> 
> How does that work with `AppendRecords` if you're adding a brand-new record? Because if I'm understanding you correctly, you'd need to set the location for that record, but since a `libdns.Address` only has `(Name, TTL, IP)` fields, there's no obvious way to do that.

This field can be omitted (use default value).

After reviewing again, I think it is OK to remove the ID field (it should not cause any problems except for some edge cases).

rrtypes.go
mholt: Hmm! Yeah I forgot about this.

I prefer simple and Go-friendly for these libraries, but I wonder, how would `Text string` represent: `"v=spf1 include:_spf.example.com ~all" "google-site-verification=abc123"`?

Would it just be `v=spf1 include:_spf.example.com ~all google-site-verification=abc123`?

libdns.go
mholt: I changed it back to ListZones :see_no_evil: Because I just realized that `GetRecords()` does actually have a parameter (the zone), and `ListZones()` is just listing all zones, with no parameters, so I don't think the symmetry feels as right.

mholt:
Sorry for my delay... been busy with family and other work stuff. (Will be spotty for the next few days as well.)

@gucci-on-fleek Thanks for your iterating on this. I think we're super close now?

> So maybe it might be better to instead define
>
> `Value    *SvcParams`
>
> That should stop it from panicking, but I don't know enough about Go to know the greater implications to making a struct field a pointer.

Hmm, that would indeed do a comparison, but it'll do a pointer comparison, nothing actually useful. It will only compare if the memory addresses are the same, not if the semantic value of the pointer is the same. So that's only useful if you want to ensure you're using the same instance/copy of the SvcParams. It doesn't check semantic meaning.

> Depending on the timeline of all of this, it might be good to have the tests ready before we notify all of the provider maintainers. Having tests would probably make it easier to port all the providers open, and it would be mildly frustrating to finish porting a provider and then see that all the tests are failing after. But [porting libdns/rfc2136 over was easy enough](https://github.com/libdns/rfc2136/commit/9787b911cbd3ec51cfe6e2ccbc5c9c6fd6463a1b) that I don't think that this should be a huge concern either way.

Good idea -- I might still want to merge this first, since I've got a big Caddy feature leaning on this (HTTPS records in particular), and then we can discuss testing in a new issue (/cc @mjl-) and then notify all the providers over time about the changes for 1.0.

As for SVCB records, I still have to review the change -- I'll probably have questions for you!

record.go
mholt: I think it sounds good and the godoc looks good concerning this now, I'll resolve this but feel free to continue discussion :+1:

rrtypes.go
```diff
@@ -0,0 +1,374 @@
...
+	// “Params” is a map of key–value pairs that are used to specify various
+	// parameters for the service. The keys are typically registered with IANA,
+	// and which keys are valid is service-dependent.
+	//
+	// Note that there is a key called “mandatory”, but this does not mean that
+	// it is mandatory for you to set the listed keys. Instead, this means that
+	// if a client does not understand all of the listed keys, then it must
+	// ignore the entire record. This is similar to the “critical” flag in CAA
+	// records.
+	Params *SvcParams
```
mholt, line 296:
```suggestion
	Params SvcParams
```

As explained, probably best to keep this a value not a pointer. (Even though it makes comparison difficult.)

rrtypes.go
```diff
@@ -0,0 +1,374 @@
...
+// and semantics.
+//
+// HTTPS-type records are  used to provide clients with information for
+// establishing HTTPS connections to servers. It may include data about ALPN,
+// ECH, IP hints, and more.
+//
+// Unlike the other RR types that are hostname-focused or service-focused, SVCB
+// (“Service Binding”) records are URL-focused. This distinction is generally
+// irrelevant, but is important when disusing the port fields.
+type SVCB struct {
```
mholt, line 188: I'm still having a hard time swallowing this as the catch-all type for SVCB, HTTPS, and potentially other types in the future.

Since this type is supposed to represent multiple record types, what if we did like you suggested with A/AAAA :arrow_right: "Address", and call this something more general like "ServiceBinding"? Like, I just feel weird knowing that I'll be working with HTTPS records, but would have to keep reasoning about this "SVCB" type.

Great godoc comments, though, really helpful for DNS noobs like me.

mholt (commented):
Did a quick once-over of the SVCB records, @gucci-on-fleek. Looks good, very thorough -- just a couple nits so far. It kind of intimidates me tbh, a bit more complicated than what I had in mind :laughing:  (but not necessarily bad... I guess most of the complexity is hidden away from the user).

rrtypes.go
gucci-on-fleek: > how would `Text string` represent: `"v=spf1 include:_spf.example.com ~all" "google-site-verification=abc123"`?
> 
> Would it just be `v=spf1 include:_spf.example.com ~all google-site-verification=abc123`?

Yes, but it is very very unlikely that you want 1 `TXT` record with the single string `v=spf1 include:_spf.example.com ~all google-site-verification=abc123` or 1 `TXT` record with the two strings `v=spf1 include:_spf.example.com ~all` and `google-site-verification=abc123`. You instead probably want 2 `TXT` records, each with a single string, `v=spf1 include:_spf.example.com ~all` and `google-site-verification=abc123` (which is what `two` does in the example above).

Here's a live example:

```console
$ cat <<EOF | nsupdate -v -p 25108 -y $RFC2136_KEYALG:$RFC2136_KEYNAME:$RFC2136_KEY
server ns.maxchernoff.ca.
zone libdns.internal. 
update delete libdns.internal.
update add 1txt-1string.libdns.internal. 0 TXT "v=spf1 include:_spf.example.com ~all google-site-verification=abc123"
update add 1txt-2string.libdns.internal. 0 TXT "v=spf1 include:_spf.example.com ~all" "google-site-verification=abc123"
update add 2txt-1string.libdns.internal. 0 TXT "v=spf1 include:_spf.example.com ~all"
update add 2txt-1string.libdns.internal. 0 TXT "google-site-verification=abc123"
send
EOF 

$ dig +noall +answer +tcp -p 25108 @ns.maxchernoff.ca libdns.internal. -y $RFC2136_KEYALG:$RFC2136_KEYNAME:$RFC2136_KEY AXFR
libdns.internal.               0  IN  SOA  ns.maxchernoff.ca. dns.maxchernoff.ca. 2025040402 0 0 0 0
1txt-1string.libdns.internal.  0  IN  TXT  "v=spf1 include:_spf.example.com ~all google-site-verification=abc123"
1txt-2string.libdns.internal.  0  IN  TXT  "v=spf1 include:_spf.example.com ~all" "google-site-verification=abc123"
2txt-1string.libdns.internal.  0  IN  TXT  "google-site-verification=abc123"
2txt-1string.libdns.internal.  0  IN  TXT  "v=spf1 include:_spf.example.com ~all"
libdns.internal.               0  IN  SOA  ns.maxchernoff.ca. dns.maxchernoff.ca. 2025040402 0 0 0 0
```

`2txt-1string` is the only correct record; with my suggestion of only allowing 1 string, `1txt-2string` becomes impossible. `1txt-1string` would still be possible (and incorrect), but to generate it with `libdns`, the user would have to manually concatenate the strings, and there's no way that we could stop them from doing that.

rrtypes.go
gucci-on-fleek: > > So maybe it might be better to instead define
> > `Value    *SvcParams`
> > That should stop it from panicking, but I don't know enough about Go to know the greater implications to making a struct field a pointer.
>
> Hmm, that would indeed do a comparison, but it'll do a pointer comparison, nothing actually useful. It will only compare if the memory addresses are the same, not if the semantic value of the pointer is the same. So that's only useful if you want to ensure you're using the same instance/copy of the SvcParams. It doesn't check semantic meaning.

Right, but I think that giving a false comparison is probably (maybe?) better than a runtime panic. Using the following test file:

```go
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/libdns/libdns"
	"github.com/libdns/rfc2136"
)

func doStuffWithRecord(rec libdns.Record) {
	fmt.Printf("Found record: %#v\n", rec)
}

func main() {
	ctx := context.Background()
	provider := rfc2136.Provider{
		KeyName: os.Getenv("RFC2136_KEYNAME"),
		KeyAlg:  os.Getenv("RFC2136_KEYALG"),
		Key:     os.Getenv("RFC2136_KEY"),
		Server:  os.Getenv("RFC2136_SERVER"),
	}
	zone := os.Getenv("RFC2136_ZONE")

	fmt.Println("Getting records from provider...")
	recs, err := provider.GetRecords(ctx, zone)
	if err != nil {
		panic(err)
	}

	fmt.Printf("Records: %#v\n", recs)

	for _, targetRecord := range recs {
		targetRecord, _ := targetRecord.RR().Parse()
		for _, rec := range recs {
			rec, _ := rec.RR().Parse()
			if rec == targetRecord {
				doStuffWithRecord(rec)
			}
		}
	}
	fmt.Println("End of records")
}
```

with `Params *SvcParams` I get

```log
Getting records from provider...
Records: []libdns.Record{libdns.RR{Name:"@", TTL:0, Type:"SOA", Data:"ns.maxchernoff.ca. dns.maxchernoff.ca. 2025040404 0 0 0 0"}, libdns.TXT{Name:"@", TTL:3600000000000, Text:"just a TXT record"}, libdns.SVCB{Scheme:"https", URLSchemePort:0x0, Name:"@", TTL:3600000000000, Priority:0x1, Target:".", Params:(*libdns.SvcParams)(0xc000124080)}, libdns.RR{Name:"@", TTL:0, Type:"SOA", Data:"ns.maxchernoff.ca. dns.maxchernoff.ca. 2025040404 0 0 0 0"}}
Found record: libdns.RR{Name:"@", TTL:0, Type:"SOA", Data:"ns.maxchernoff.ca. dns.maxchernoff.ca. 2025040404 0 0 0 0"}
Found record: libdns.RR{Name:"@", TTL:0, Type:"SOA", Data:"ns.maxchernoff.ca. dns.maxchernoff.ca. 2025040404 0 0 0 0"}
Found record: libdns.TXT{Name:"@", TTL:3600000000000, Text:"just a TXT record"}
Found record: libdns.RR{Name:"@", TTL:0, Type:"SOA", Data:"ns.maxchernoff.ca. dns.maxchernoff.ca. 2025040404 0 0 0 0"}
Found record: libdns.RR{Name:"@", TTL:0, Type:"SOA", Data:"ns.maxchernoff.ca. dns.maxchernoff.ca. 2025040404 0 0 0 0"}
End of records
```

but with `Params SvcParams` I get

```log
Getting records from provider...
Records: []libdns.Record{libdns.RR{Name:"@", TTL:0, Type:"SOA", Data:"ns.maxchernoff.ca. dns.maxchernoff.ca. 2025040404 0 0 0 0"}, libdns.TXT{Name:"@", TTL:3600000000000, Text:"just a TXT record"}, libdns.SVCB{Scheme:"https", URLSchemePort:0x0, Name:"@", TTL:3600000000000, Priority:0x1, Target:".", Params:libdns.SvcParams{"alpn":[]string{"h2", "h3"}}}, libdns.RR{Name:"@", TTL:0, Type:"SOA", Data:"ns.maxchernoff.ca. dns.maxchernoff.ca. 2025040404 0 0 0 0"}}
Found record: libdns.RR{Name:"@", TTL:0, Type:"SOA", Data:"ns.maxchernoff.ca. dns.maxchernoff.ca. 2025040404 0 0 0 0"}
Found record: libdns.RR{Name:"@", TTL:0, Type:"SOA", Data:"ns.maxchernoff.ca. dns.maxchernoff.ca. 2025040404 0 0 0 0"}
Found record: libdns.TXT{Name:"@", TTL:3600000000000, Text:"just a TXT record"}
panic: runtime error: comparing uncomparable type libdns.SVCB

goroutine 1 [running]:
main.main()
        /home/max/Projects/libdns-rfc2136/_nogit-testing/main.go:38 +0x4c5
exit status 2
```

Now, This is admittedly a contrived example, so maybe this isn't relevant in regular code?

rrtypes.go
gucci-on-fleek: > I'm still having a hard time swallowing this as the catch-all type for SVCB, HTTPS, and potentially other types in the future.

Yeah, I wish that they just made a single `SVCB` RR type, instead of defining a `HTTPS` RR type that's almost the same as `SVCB` but not quite. 

> Since this type is supposed to represent multiple record types, what if we did like you suggested with A/AAAA ➡️ "Address", and call this something more general like "ServiceBinding"?

Excellent idea!

gucci-on-fleek:
> Sorry for my delay... been busy with family and other work stuff. (Will be spotty for the next few days as well.)

No problem at all.

> I think we're super close now?

I think so too.

> > Depending on the timeline of all of this, it might be good to have the tests ready before we notify all of the provider maintainers. Having tests would probably make it easier to port all the providers open, and it would be mildly frustrating to finish porting a provider and then see that all the tests are failing after. But [porting libdns/rfc2136 over was easy enough](https://github.com/libdns/rfc2136/commit/9787b911cbd3ec51cfe6e2ccbc5c9c6fd6463a1b) that I don't think that this should be a huge concern either way.
>
> Good idea -- I might still want to merge this first, since I've got a big Caddy feature leaning on this (HTTPS records in particular), and then we can discuss testing in a new issue (/cc @mjl-) and then notify all the providers over time about the changes for 1.0.

:+1:

rrtypes.go
mholt: > 2txt-1string is the only correct record

You mean, for SPF & site verification specifically, or syntactically? I ask because obviously there's lots of uses of TXT records outside of SPF & site verification, but I don't know of any common ones that require 1txt-2string.

So, maybe this is where the "we go for 99.9% of use cases" comes in. I just don't know how common it is. I assume it's in that 0.1% or less. Way less.

> 1txt-1string would still be possible (and incorrect), but to generate it with libdns, the user would have to manually concatenate the strings, and there's no way that we could stop them from doing that.

But, 2txt-1string is possible and easy, right? If we use `Text string`? I'm OK with it if "there's no way we could stop them from doing that", I suppose, it just wouldn't work once they apply the changes. I guess I don't know a better API; so far I'm still leaning towards `Text string`. Thoughts?

rrtypes.go
mholt: Good point, that code _looks_ like it should work. (But `RR().Parse()` is a bit of a weird pattern, yeah?)

Thinking on this more, it seems like a bad idea to try to allow primitive comparison of `Record` values.

Extending my argument about `SvcParams` not being obviously comparable more broadly, is it reasonable to suggest that equality of an entire, arbitrary `Record` (whatever type it is) isn't defined/obvious?

`RR` yes, but arbitrary `Record` values, no, I don't believe so...

So, maybe this is a good thing to bring up in the docs: `Record` values shouldn't be used for comparison. It's not obvious what equality means in terms of a DNS record, and it may be different for everyone.

If people really wanted to compare records primitively, I think they can compare the RR structs. Otherwise I think it's best that they write their own comparison method.

mholt:
Alrighty, I think pending the resolution of these two discussions:

- Record comparison and equality (whether to support `==`, and then document expectations)
- TXT layout (`Text string` or `Text []string`)

We can probably merge this in, do a `v1.0.0-beta.1` release, and then discuss testing in another issue.

rrtypes.go
gucci-on-fleek: > > 2txt-1string is the only correct record
> 
> You mean, for SPF & site verification specifically, or syntactically?

Yes, for SPF and site verification specifically.

> there's lots of uses of TXT records outside of SPF & site verification, but I don't know of any common ones that require 1txt-2string.

Based off of the [DNSControl link](https://docs.dnscontrol.org/developer-info/opinions#opinion-8-txt-records-are-one-long-string), I don't think that there are _any_ that require 1txt-2string (but it's hard to prove a negative of course).

> So, maybe this is where the "we go for 99.9% of use cases" comes in. I just don't know how common it is. I assume it's in that 0.1% or less. Way less.

:+1:

> > 1txt-1string would still be possible (and incorrect), but to generate it with libdns, the user would have to manually concatenate the strings, and there's no way that we could stop them from doing that.
> 
> But, 2txt-1string is possible and easy, right? 

Yes, you do that like you would with adding two different addresses to a website, by adding two separate records for the same name:

```go
provider.SetRecords(ctx, zone, []libdns.Record{
	libdns.TXT{
		Name: "@",
		TTL:  time.Hour,
		Text: "v=spf1 include:_spf.example.com ~all",
	},
	libdns.TXT{
		Name: "@",
		TTL:  time.Hour,
		Text: "google-site-verification=abc123",
	},
})
```

>  guess I don't know a better API; so far I'm still leaning towards `Text string`. Thoughts?

Yeah, I think that `Text string` (what I have right now) makes the most sense too.

rrtypes.go
gucci-on-fleek: > But `RR().Parse()` is a bit of a weird pattern, yeah?

Hmm, now that you mention it, that might actually be necessary in a lot of cases right now. It would be valid for an implementation to do something like return `libdns.Address` structs for `A`/`AAAA` RRs and `libdns.Record` for all other record types, so if you want to make sure that you always have a specific record type (for a type switch), you'd have to run `RR().Parse()` first. 

Should we maybe specify that `libdns` implementations should always return the most specific record type possible?

> Extending my argument about `SvcParams` not being obviously comparable more broadly,

In the DNS (binary) wire format, the `SvcParams` keys are required to be increasing numeric order of their numeric value ([RFC 9460 §2.2](https://datatracker.ietf.org/doc/html/rfc9460#section-2.2-5), so there is a canonical/official way to compare equality. But if we wanted to implement this, we'd need to hard-code in the [list of valid keys](https://www.iana.org/assignments/dns-svcb/dns-svcb.xhtml) to make sure that we sort them properly.

> It's not obvious what equality means in terms of a DNS record, and it may be different for everyone.

Because DNSSEC signs the hash of a record, each RR type has an official/canonical representation, so I'd say that that would be the “obvious” definition of equality, but it would be a lot of work for us to implement those semantics.

> If people really wanted to compare records primitively, I think they can compare the RR structs. Otherwise I think it's best that they write their own comparison method.

:+1:, I think that this is the best solution.

> Thinking on this more, it seems like a bad idea to try to allow primitive comparison of `Record` values.
>
> [...]
>
> So, maybe this is a good thing to bring up in the docs: `Record` values shouldn't be used for comparison.

Ok, I'll add this to the documentation. What should we do with the `Params` field though? I would still lean towards `Params *SvcParams` because otherwise, if `libdns` implementations ignore the documentation and primitively compare `Record` values, we could end up in a place where everything works fine until someone adds a `HTTPS` RR to the DNS and then their program starts panicking. 

For a worst-case scenario, someone could enable ECH in Caddy, which then adds an `HTTPS` RR, which then makes the `libdns` implementation panic, which then crashes Caddy. But then even if they disabled ECH after the first crash, the `HTTPS` RR would still be there, so Caddy would keep crashing.

But that being said, the comparison with `Params *SvcParams` would be bogus, so maybe it would make more sense to use `Params SvcParams` so that the error would be obvious?

rrtypes.go
mholt: Awesome, alright -- sounds good. Let's stick with `Text string` then.

rrtypes.go
mholt: > Hmm, now that you mention it, that might actually be necessary in a lot of cases right now. It would be valid for an implementation to do something like return libdns.Address structs for A/AAAA RRs and libdns.Record for all other record types, so if you want to make sure that you always have a specific record type (for a type switch), you'd have to run RR().Parse() first.

Ahh, good point.

So I had a big long reply drafted up to this, documenting my thought process that you spurred on, but then my browser tab closed and I lost it.

Basically... yeah, let's document that providers should return specific type structs rather than RR.

There's just no way to technically enforce and require that they do without being evil:

```go
// remove RR() from the RR struct, then:

type Record interface {
    RR() RR
    mustUseOurTypes() unexportedType
}
```

This is the only way I can think of that would prevent them from defining their own types that implement the Record interface (making type assertions difficult again), and force that they don't return `RR` structs.

(I also like the flexibility (for input purposes) of RR also being a Record.)

mholt:
Okay, awesome!! Thanks again for the reviews and discussion!

> Because DNSSEC signs the hash of a record, each RR type has an official/canonical representation, so I'd say that that would be the “obvious” definition of equality, but it would be a lot of work for us to implement those semantics.

We'll see what is requested by users after we merge this, we can always add this later, I suppose.

I think with my latest comments, if you are good with them, we have resolved the remaining issues.

I'm ready to merge whenever. Just need an approving review.

mholt:
I went ahead and changed *SvcParams to SvcParams -- I hope that's okay.

**I'm ready for a final review and approval, if possible. The code in its current state has my approval (but I can't review it since it's my own PR).**

Thank you again for the collaboration!

We can still make changes after merging this, but I'm ready to merge this and have a few providers start to upgrade.

mholt:
Oops, one last change :smile: I had forgot to document what we decided, that input should be able to be any `Record` value, but outputs should be specific types so type-switches can succeed easily and without errors.

rrtypes.go
gucci-on-fleek: > I see what you're saying, but, pointers to maps are quite unconventional. I also think it's generally unwise to primitively compare interface types. At least the panic alerts you to incorrect code, but using a pointer will just hide the bug. I think it should not be a pointer.

Ok, that totally works for me. I think that I agree with you now that the panic is actually good since it lets you know about the error right away instead of hiding it somewhere in the code. Thanks for hearing me out though.

gucci-on-fleek:
> > Because DNSSEC signs the hash of a record, each RR type has an official/canonical representation, so I'd say that that would be the “obvious” definition of equality, but it would be a lot of work for us to implement those semantics.
> 
> We'll see what is requested by users after we merge this, we can always add this later, I suppose.

Oh yeah, I definitely agree that there's no need to add this now (or ever, maybe).

> I went ahead and changed *SvcParams to SvcParams -- I hope that's okay.

Yup.

> **I'm ready for a final review and approval, if possible. The code in its current state has my approval (but I can't review it since it's my own PR).**
> 
> Thank you again for the collaboration!
> 
> We can still make changes after merging this, but I'm ready to merge this and have a few providers start to upgrade.

:+1:

gucci-on-fleek (approved):

mholt:
Heck yeah, thank you everyone, for your contributions! This will soon be tagged as v1.0.0-beta.1.

mholt:
@gucci-on-fleek @mjl- @devhaozi FYI, I've submitted an issue to the top several DNS provider packages (according to the Caddy download page) to get their attention on this ASAP, so their experience can help inform us if we need to make any further breaking changes during the beta period. After a little time (probably still during the beta period) we'll open the firehose and ask the others to also upgrade.

devhaozi:
> @gucci-on-fleek @mjl- @devhaozi FYI, I've submitted an issue to the top several DNS provider packages (according to the Caddy download page) to get their attention on this ASAP, so their experience can help inform us if we need to make any further breaking changes during the beta period. After a little time (probably still during the beta period) we'll open the firehose and ask the others to also upgrade.

The two packages https://github.com/libdns/westcn and https://github.com/libdns/huaweicloud that I maintain have been upgraded to v1.0.0-beta.1, and there is no problem.

bsutton:
this change has broken the WeidiDeng/caddy-cloudflare-ip.

RUN xcaddy build \
    --with github.com/caddy-dns/cloudflare \
    --with github.com/WeidiDeng/caddy-cloudflare-ip

https://github.com/WeidiDeng/caddy-cloudflare-ip/issues/9

mholt:
Yes, just build with the prior commit until other packages update.

JMyklebust:
Working on updating https://github.com/libdns/domainnameshop for this and made good headway.
However the biggest thing I'm missing from the old is the ``ID`` field on ``RR``....

There is a need for a unique identifier that can be kept track of, ``Name`` is not good enough since it's not a unique field.

In the case of Domanname.shop the API returns an ID when you create a record and you need an ID to remove a record.
And querying for a specific record is also based on this ID.

Without an ID field in the ``RR`` type I have to keep track of this ID some other way since I can't expect it to be carried by a ``libdns.Record``.
Or I have to query the entire zone all the time and then compare the name and data and guess that it matches.

devhaozi:
> Working on updating https://github.com/libdns/domainnameshop for this and made good headway. However the biggest thing I'm missing from the old is the `ID` field on `RR`....
> 
> There is a need for a unique identifier that can be kept track of, `Name` is not good enough since it's not a unique field.
> 
> In the case of Domanname.shop the API returns an ID when you create a record and you need an ID to remove a record. And querying for a specific record is also based on this ID.
> 
> Without an ID field in the `RR` type I have to keep track of this ID some other way since I can't expect it to be carried by a `libdns.Record`. Or I have to query the entire zone all the time and then compare the name and data and guess that it matches.

See https://github.com/libdns/libdns/pull/153#discussion_r2023368283

JMyklebust:
That's a bit of hassle, but I've made it work and updated my module, got it working with Caddy which is my main focus.
Though some edge cases of multiple "identical"/similar records is not cleanly solved.

mholt:
@JMyklebust Thanks for the feedback / update! Yeah, the ID field was a tricky one... At first it seemed obvious we should have it, but then there's some very subtle and potentially harmful outcomes of doing that, considering cross-zone and cross-provider compatibility... anyway you probably read the discussion and get the gist.

Thanks for maintaining your package -- changes look good, at a quick glance :+1:

Issue #154, opened 2025-03-23
Rate limitting, atomic transactions, and batching


gucci-on-fleek:
/cc @mholt 

A few semi-related thoughts:

### Rate limitting

Essentially every provider will have a rate limit of some sort. If you hit the rate limit, should the provider catch that itself and automatically retry, or should it just return an error and make that the user's problem?

The provider is probably in a better place to handle this since it can catch a `429 Too Many Requests` and wait specifically until `Retry-After`; however, if the API that the provider calls doesn't send a `Retry-After`, then the provider will have to do exponential backoff (or something similar), which could block for a very long time and potentially surprise any downstream users.

### Atomic transactions

Should we guarantee that `SetRecords` is atomic/a complete transaction? That is, if `err == nil`, then all of the changes in `SetRecords` were applied; and if `err != nil`, then calling `SetRecords` didn't change the zone at all.

Many providers will need to loop over the records passed to `SetRecords` and make multiple individual API calls to set the records. If the provider fails to set one of these records, should we require it to go back and revert all of the changes that it made in earlier loop iterations? And what should it do if it gets an error while reverting a change?

I think that having `SetRecords` be atomic is the only sensible solution, since otherwise `err != nil` means that the zone is in a completely unknown state. But implementing these semantics would be pretty tricky for some `libdns` providers if the upstream APIs that they call don't provide the necessary primitives.

### Batched updates

Should we add a function that allows you to collect multiple updates in a single batch? Right now, you can combine multiple `SetRecords` and `AppendRecords` calls into a single `SetRecords` call (possibly preceded by a `GetRecords` call), but there's no way to combine this with any `DeleteRecords` calls.

Without a method that lets you batch multiple actions into a single atomic transaction, then downstream users would need code like this:

```go
// Assume the zone begins with the following contents:
//
//	alpha  3600  IN  A      192.0.2.111
//	beta   3600  IN  A      192.0.2.222
//	www    3600  IN  CNAME  alpha
//
// and you want for the final contents to be:
//
//	beta  3600  IN  A  192.0.2.111
//	www   3600  IN  A  192.0.2.111
//	www   3600  IN  A  192.0.2.222
//
// You want this to either fully succeed or fully fail.
func AtomicSetZone() error {
	provider := &Provider{}
	ctx := context.Background()
	zone := "example.com."

	records, err := provider.GetRecords(ctx, zone)
	if err != nil {
		return err
	}

	_, err = provider.DeleteRecords(ctx, zone, []Record{
[...]

PR #156, opened 2025-03-25
Updates to #153

gucci-on-fleek:
As discussed in https://github.com/libdns/libdns/pull/153#discussion_r2010592489 and #154.


mholt (changes requested):
Looks great to me! Thanks for filling out 3 more record types!

Just one nit on a comment but not a big deal.

PR #164, opened 2025-04-18
Support 2-label SRV names (fix #163)

### LINKED ISSUE #163: SRV records for the zone itself don't parse with RR.Parse
IIUC,

```
$ORIGIN mydomain.tld.
_sip._tcp IN SRV 1 100 5061 sip.example.com.
```

is represented in libdns v1.0.0-beta.1 as

```go
libdns.RR{
	Name: "_sip._tcp",
	Type: "SRV",
	TTL:  1 * time.Hour,
	Data: "1 100 5061 sip.example.com.",
}
```

However, using calling RR.Parse fails:

```go
_, err := libdns.RR{
	Name: "_sip._tcp",
	Type: "SRV",
	TTL:  1 * time.Hour,
	Data: "1 100 5061 sip.example.com.",
}.Parse()
fmt.Println(err) //  name _sip._tcp does not contain enough fields; expected format: '_service._proto.name'
```

https://go.dev/play/p/NQJpC52cJQw

mholt:
@gucci-on-fleek You know a bit more about this than I do, does this look right?

### DIFF
--- a/record.go
+++ b/record.go
@@ -299,6 +299,14 @@ func (r RR) toServiceBinding() (ServiceBinding, error) {
 	scheme := ""
 	var port uint64 = 0
 	nameParts := strings.SplitN(r.Name, ".", 3)
+	// Handle the case where the name is only underscore-prefixed labels
+	if len(nameParts) <= 1 && strings.HasPrefix(nameParts[0], "_") {
+		nameParts = append(nameParts, "@")
+	} else if len(nameParts) == 2 && strings.HasPrefix(nameParts[1], "_") {
+		nameParts = append(nameParts, "@")
+	}
+
+	// Parse the first two parts of the name
 	if strings.HasPrefix(nameParts[0], "_") && strings.HasPrefix(nameParts[1], "_") {
 		portStr := strings.TrimPrefix(nameParts[0], "_")
 		scheme = strings.TrimPrefix(nameParts[1], "_")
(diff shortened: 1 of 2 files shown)


record.go
```diff
@@ -251,14 +251,18 @@ func (r RR) toSRV() (SRV, error) {
...
 	parts := strings.SplitN(r.Name, ".", 3)
-	if len(parts) < 3 {
-		return SRV{}, fmt.Errorf("name %v does not contain enough fields; expected format: '_service._proto.name'", r.Name)
+	if len(parts) < 2 {
+		return SRV{}, fmt.Errorf("name %v does not contain enough fields; expected format: '_service._proto.name' or '_service._proto'", r.Name)
+	}
+	name := "@"
+	if len(parts) == 3 {
+		name = parts[2]
 	}
```
znkr, line 260: One more edge case is a two part name ending with `.`. For example, `_sip._tcp.`. That's not a valid input I think, but it would parse as `Name: ""`.

gucci-on-fleek: I think that `_sip._tcp.` falls into the case of “semantically valid, but mostly nonsensical”. There's nothing technically preventing the root zone from hosting an SIP server, although it would be pretty weird if it did. So I think that either allowing or disallowing this would both be reasonable.

gucci-on-fleek (approved):
LGTM. I also noticed that `ServiceBinding`s had the same issue, so I pushed some (rather ugly) code to fix this there too.

record.go
mholt: Let's cross that bridge when we come to it (i.e. someone actually has that problem). Thanks!

Issue #161, opened 2025-04-16
ServiceBinding weirdness


znkr:
I am in the process of upgrading the desec provider and ran into a snag when I tried to implement support for HTTP records. The biggest problem I have is that I don't really know how to deconstruct the query name into the different pieces of a ServiceBinding. That is, parsing the name and then understanding what piece of the name represents the port and what represents the name is not trivial. I think I am going to work around this by returning a wrapped RR type for now.

Also, I think it's confusing that when getting a record for "_1234._https.foo.bar" ends up returning a record with a different name.

znkr:
Never mind, I just discovered the RR.Parse method

mholt:
Yeah the HTTPS (and in general, all Service Binding -- and SRV) record types are confusing because they kind of use a hack to encode additional data into the "name" field... really unfortunate IMO but I'm sure they had reasons for it.

znkr:
I haven't thought too much about this, but to me it feels a bit off to parse the port out of the name. After all, you probably wouldn't do this for acme TXT records either. I really don't want to reopen the discussion though. :)

mholt:
> Also, I think it's confusing that when getting a record for "_1234._https.foo.bar" ends up returning a record with a different name.

To clarify, it's the _parsed_ (interpreted) name; the `RR` (available with `.RR()`) has it as you would expect in its full form. Both can be useful, we just had to choose one.

gucci-on-fleek:
> I really don't want to reopen the discussion though.

Oh no, no problem there at all.

> but to me it feels a bit off to parse the port out of the name.

I agree that it's a little bit weird, but it makes for a nice symmetry where the `Name`s are all the same for any given hostname:

```go
[]libdns.Record{
	libdns.Address{
		Name: "www",
		TTL:  300 * time.Second,
		IP:   netip.MustParseAddr("192.0.2.1"),
	},
	libdns.Address{
		Name: "www",
		TTL:  300 * time.Second,
		IP:   netip.MustParseAddr("2001:db8::1"),
	},
	libdns.ServiceBinding{
		Name:     "www",
		TTL:      300 * time.Second,
		Scheme:   "https",
		Priority: 1,
		Target:   ".",
	},
	libdns.ServiceBinding{
		Name:     "www",
		TTL:      300 * time.Second,
		Scheme:   "dns",
		URLSchemePort: "853",
		Priority: 1,
		Target:   ".",
	},
	libdns.SRV{
		Name:      "www",
[...]

PR #165, opened 2025-04-19
Don't use `.@` for relative names in [SRV|ServiceBinding].RR()

gucci-on-fleek:
While replying to #161, I noticed that `.RR()` gives odd results for `SRV` and `ServiceBinding` records at the zone apex:

```go
package main

import (
	"fmt"
	"time"

	"github.com/libdns/libdns"
)

func main() {
	records := []libdns.Record{
		libdns.ServiceBinding{
			Name:          "@",
			Scheme:        "examplescheme",
			URLSchemePort: 1234,
			TTL:           1 * time.Hour,
			Priority:      1,
			Target:        ".",
			Params:        libdns.SvcParams{},
		},
		libdns.SRV{
			Name:      "@",
			Service:   "exampleservice",
			Transport: "tcp",
			TTL:       1 * time.Hour,
			Priority:  1,
			Weight:    2,
			Target:    ".",
		},
	}

	rrs := make([]libdns.RR, len(records))
	for i, record := range records {
		rrs[i] = record.RR()
	}

	for _, rr := range rrs {
		fmt.Println(rr.Name)
	}
}
```
```log
_1234._examplescheme.@
_exampleservice._tcp.@
```

`libdns` has no problem parsing names like `_1234._examplescheme.@`, but users would typically expect `_1234._examplescheme` instead, so I've updated the code to trim any `.@` suffixes.

### DIFF
--- a/record_test.go
+++ b/record_test.go
@@ -477,3 +477,64 @@ func TestSvcParamsString(t *testing.T) {
 		}
 	}
 }
+
+func TestRelativeRRNames(t *testing.T) {
+	for _, test := range []struct {
+		input  Record
+		expect string
+	}{
+		{
+			input: ServiceBinding{
+				Name:          "@",
+				Scheme:        "examplescheme",
+				URLSchemePort: 1234,
+				TTL:           1 * time.Hour,
+				Priority:      1,
+				Target:        ".",
+				Params:        SvcParams{},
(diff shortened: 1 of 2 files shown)


mholt (approved):
Thanks! Nice catch.

PR #166, opened 2025-04-19
Make all zero values have `.RR().Data == ""`

gucci-on-fleek:
(Noticed in https://github.com/libdns/rfc2136/pull/9#discussion_r2051132593)

The `Address.IP` field is allowed to be zero-initialized for use in `DeleteRecords`:

https://github.com/libdns/libdns/blob/5ab1d4de259f1eb914085c61784ad9176ea8e803/libdns.go#L200-L205

Currently, calling `.RR()` on such an `Address` gives `rr.Data == "invalid IP"`, but this is a fairly surprising zero value, so this PR changes this to use the more conventional `""`.

### DIFF
--- a/rrtypes.go
+++ b/rrtypes.go
@@ -28,11 +28,19 @@ func (a Address) RR() RR {
 	if a.IP.Is6() {
 		recType = "AAAA"
 	}
+	data := a.IP.String()
+	if a.IP == (netip.Addr{}) {
+		// If the IP address is null, then we get the string "invalid IP". We'll
+		// convert this to the empty string to make
+		// [libdns.RecordDeleter.DeleteRecords] easier to use when missing IP
+		// addresses are passed.
+		data = ""
+	}
 	return RR{
 		Name: a.Name,
 		TTL:  a.TTL,
 		Type: recType,
-		Data: a.IP.String(),
+		Data: data,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


gucci-on-fleek:
Hmm, I suppose that this really should apply to the zero values for all `Record`s, not just `Address`es.

rrtypes.go
```diff
@@ -49,11 +57,16 @@ type CAA struct {
 }
 
 func (c CAA) RR() RR {
+	data := fmt.Sprintf(`%d %s %q`, c.Flags, c.Tag, c.Value)
+	// Make sure that the zero value is an empty string
+	if data == `0  ""` {
```
mholt, line 62:
```suggestion
	data := fmt.Sprintf(`%d %s %q`, c.Flags, c.Tag, c.Value)
	// Make sure that the zero value is an empty string
	if c.Flags == 0 && c.Tag == "" && c.Value == "" {
```
This feels better to me since it doesn't rely on the exact syntactic combination of the `data` variable. But I guess either could work, I dunno. This also feels a little clearer. (And maybe for the other RR types too -- but this is just a nit, I guess it doesn't really matter)

mholt (approved):
Thanks for this! I hadn't thought about this :sweat_smile: I did have one nit, but it's totally bikeshedding. Just wanted your thoughts on it. I'm good to merge it either way.

rrtypes.go
gucci-on-fleek: No, that seems like a _much_ better way of doing this; good idea!

gucci-on-fleek:
Ok, I've rebased onto `master` and fixed the `data` checks. Can you please tag a `-beta.2` once this is merged so that I can use it at [libdns/rfc2136](https://github.com/libdns/rfc2136)? Thanks again.

mholt (approved):
LGTM! Thank you!!

PR #169, opened 2025-04-21
Create RR.ProviderData (close #119)

### LINKED ISSUE #119: How to pass additional configs of record, typically cloudflare's proxied option
looks like it's been ignored
https://github.com/libdns/cloudflare/blob/master/models.go#L112-L132

mholt:
This is my proposal for being able to attach provider-specific (meta?)data to an RR. See #119.

However, since provider implementations are supposed to return the concrete struct types, not RRs, we'll likely need to add this field to all the specific structs (Address, TXT, CNAME, etc...).

I also added JSON tags to the RR struct, in preparation for 1.0, since JSON-serialization seems like it could be a common pattern. (Caddy/CertMagic do not do this, but I could see having separate JSON identifiers being handy.)

Finally, I am torn between `Meta`, `Metadata`, and `ProviderData` for the field name. Any thoughts?

@gucci-on-fleek What do you think?

### DIFF
--- a/libdns.go
+++ b/libdns.go
@@ -19,6 +19,15 @@
 // Specific record types such as [Address], [SRV], [TXT], and others implement
 // the [Record] interface.
 //
+// Record values should not be primitvely compared (==) unless it is [RR],
+// because some struct types contain maps, for which equality is not defined;
+// additionally, some packages may attach custom data to each RR struct-type's
+// `ProviderData` field, which values might not be comparable either. The
+// `ProviderData` fields are not portable across providers, or possibly even
+// zones. Because it is not portable, and we want to ensure that [RR] structs
+// remain both portable and comparable, the `RR()` methods do not preserve
+// `ProviderData` in their return values.
+//
 // Implementations of the libdns interfaces should accept as input any [Record]
 // value, and should return as output the concrete struct types that implement
 // the [Record] interface (i.e. [Address], [TXT], [ServiceBinding], etc). This
--- a/libdns.go
+++ b/libdns.go
@@ -22,7 +22,7 @@
 // Record values should not be primitvely compared (==) unless it is [RR],
 // because some struct types contain maps, for which equality is not defined;
 // additionally, some packages may attach custom data to each RR struct-type's
-// `ProviderData` field, which values might not be comparable either. The
+// `ProviderData` field, whose values might not be comparable either. The
 // `ProviderData` fields are not portable across providers, or possibly even
 // zones. Because it is not portable, and we want to ensure that [RR] structs
 // remain both portable and comparable, the `RR()` methods do not preserve
--- a/record.go
+++ b/record.go
@@ -103,13 +103,6 @@ type RR struct {
 	// Implementations are not expected to support RFC 3597 “\#” escape
 	// sequences, but may choose to do so if they wish.
 	Data string `json:"data"`
-
-	// Optional data associated with the provider serving this RR.
-	// This could be the data structure the provider's API returns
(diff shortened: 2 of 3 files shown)


mholt:
(Super skeptical of an AI review but I'm curious, so...)

pcfreak30:
seems simple enough, and btw AI is good enough to find issues (I use it near daily), but very far away from replacing us :P.

@mholt i would add `omitempty` and i think this is good.

mholt:
> i would add omitempty and i think this is good.

Thanks; I think only `ProviderData` (as currently named) would ever be empty... I think? Maybe TTL, depending on when/where provider defaults are applied.

gucci-on-fleek:
> I also added JSON tags to the RR struct, in preparation for 1.0, since JSON-serialization seems like it could be a common pattern.

:+1:

> Finally, I am torn between `Meta`, `Metadata`, and `ProviderData` for the field name. Any thoughts?

`ProviderData` seems like the least-worst name that I can think of. It's not an amazing name, but certainly better than anything else that I can think of.

> However, since provider implementations are supposed to return the concrete struct types, not RRs, we'll likely need to add this field to all the specific structs (Address, TXT, CNAME, etc...).

Hmm, yeah, you're probably right here. So then calling `.Parse()` or `.RR()` should probably just copy over the contents of `ProviderData` unmodified? I guess the other options would be to provide a hook for this or unconditionally set `ProviderData` to `nil`, but neither of those seem great to me.

Or, maybe it would be better to _only_ have `ProviderData` on the specific structs, and not on `RR`? My thinking being that you can change the `Type` of an `RR`, and providers might handle that very poorly. I don't think that this should be very much of a problem though, just something to think about.

> @gucci-on-fleek What do you think?

The main problem that I can see is that we'll either need to document that `RR`s aren't comparable

https://github.com/libdns/libdns/blob/10a2b0b2c222c51900827339aaf1e22f5cfd2528/record.go#L13-L15

which might be annoying since that means that there's no “easy” way to compare records, or document that `ProviderData` must be primitively-comparable, which would also be annoying since then you couldn't use it to hold a `map`.

The other problem that I can think of is that weird things might happen if you try and copy `Record`s between different providers, so we should probably document either that `libdns` implementations must ignore invalid/unknown `ProviderData`s, or that users must set `ProviderData` to `nil` when copying between zones/providers.

But overall, I'm very much in favour of this.

mholt:
@gucci-on-fleek Okay, I reworked it so that ProviderData is in each individual struct type, not the RR struct. I also added some package-level godoc surrounding it.

Since packages are supposed to return the individual struct types anyway, they can just attach their custom data in this field, and callers' type assertions will still work. Obviously, the provider-specific data is lost with `RR()` (and thus, `Parse()` does not populate it), but that's OK since the `RR` type is meant to be more generic/portable, and comparable.

libdns.go
```diff
@@ -19,6 +19,15 @@
 // Specific record types such as [Address], [SRV], [TXT], and others implement
 // the [Record] interface.
 //
+// Record values should not be primitvely compared (==) unless it is [RR],
+// because some struct types contain maps, for which equality is not defined;
+// additionally, some packages may attach custom data to each RR struct-type's
+// `ProviderData` field, which values might not be comparable either. The
```
gucci-on-fleek, line 25:
```suggestion
// `ProviderData` field, whose values might not be comparable either. The
```

gucci-on-fleek (approved):
LGTM, hopefully this helps make it easier to port implementations to v1.0.

mholt:
Awesome, thanks!

Issue #171, opened 2025-04-25
Simply.com provider.


alphaleonis:
Implement a provider supporting simply.com.

(I'm working on it)

PR #172, opened 2025-04-25
Add Simply.com provider.

alphaleonis:
A provider for Simply.com.

See #171


mholt:
Thank you! Looks good at a quick glance. I've given you access to the new repos so you can push to them.

alphaleonis:
Thanks @mholt, I got an invite which I accepted but I don't seem to have access to push to the new repo?

mholt:
@alphaleonis I'm not sure; you definitely have "maintain" role on both the libdns and caddy-dns repos, so you can push to it if your local git is configured properly I guess.

PR #174, opened 2025-05-04
Clarify `SetRecords` example/doc

NicolaiSoeborg:
From my understanding the example is wrong, SetRecords should "overwrite" _::3_ and _::4_ (like it does for the _A_ record above)

### DIFF
--- a/libdns.go
+++ b/libdns.go
@@ -177,20 +177,22 @@ type RecordSetter interface {
 	// Example 2:
 	//
 	//	;; Original zone
-	//	a.example.com. 3600 IN AAAA 2001:db8::1
-	//	a.example.com. 3600 IN AAAA 2001:db8::2
-	//	b.example.com. 3600 IN AAAA 2001:db8::3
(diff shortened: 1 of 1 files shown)


gucci-on-fleek (changes requested):
No, this is correct as-is; calling `SetRecords` only on records with names `a.example.com.` should not affect records the name `b.example.com.`. But I do see how you could easily miss that the first three lines are for a different domain than the last two lines since there's only a single-character difference between them.

Could you please replace the names `a.example.com.` and `b.example.com.` with either `one.example.com.` and `two.example.com.` or `alpha.example.com.` and `beta.example.com.` (your choice). That should hopefully make this clearer for future users. Thanks!

NicolaiSoeborg:
Ahh yep completely missed the a/b part! Thanks for pointing that out.
Updated with alpha/beta to make it more visible

gucci-on-fleek (approved):
Thanks!

mholt (approved):
Thanks both! I like legibility. :nerd_face:

PR #176, opened 2025-05-06
Introduce AtomicErr (close #175)

### LINKED ISSUE #175: RecordSetter isn't atomic in practice
The spec indicates that updates are supposed to be atomic, and if they cannot be atomic, then they must be rolled back. If even that is difficult, then non-conformance must be documented.

What seems to happen in practice is that updates are usually non-atomic, rollback is never attempted and this non-conformance is often not documented. Return values seem up in the air, some return the records that were modified along with an error, others return no records (even though some were modified) and an error.

Going through the providers one by one:

*  [namesilo](https://github.com/libdns/namesilo/blob/master/provider.go#L59-L88) - ❌  Updates individual records, no rollback attempt, no documentation, returns modified records and error on failure
* [cloudflare](https://github.com/libdns/cloudflare/blob/b46a2b0885f6701e1b2d930067e338eee8876e27/provider.go#L127-L184) - ❌ Updates individual records, no rollback attempt, no documentation, returns nil and error on failure
* [westcn](https://github.com/libdns/westcn/blob/a0dc1111d880db2fd20d2e325005d00e9903e5d1/provider.go#L69-L102) - ❌  Updates individual records, no rollback attempt, no documentation, returns nil and error on failure. It's worth pointing out in this case that updates are not permitted, so they must delete first and replace the record, so if creating the new record fails, it can leave the zone in an unexpected state.
* [inwx](https://github.com/libdns/inwx/blob/63c6db833546356b9cad2a684b2c36c8e78f1d9e/provider.go#L85-L134) - ❌  Updates individual records, no rollback attempt, no documentation, returns nil and error on failure
* [rfc2136](https://github.com/libdns/rfc2136/blob/5077eb0489efc099df14892e46e109b510819d5a/provider.go#L94-L117) - ✅  Conforms, as one would expect
* [huaweicloud](https://github.com/libdns/huaweicloud/blob/128b4737c9e446f674236b8b1d7e906fc549f0b8/provider.go#L70-L115) - ❌ Updates individual records, no rollback attempt, no documentation, returns nil and error on failure
* [simplydotcom](https://github.com/libdns/simplydotcom/blob/dd553b4fa0f389eb3f6db5da650896b79d97cb7f/provider.go#L130-L174) - 🟨 Updates individual records, no rollback attempt, *has* documentation, returns nil and error on failure
* [dynu](https://github.com/libdns/dynu/blob/8908f99614e1b611ac87f5288e7c35fa4107e325/provider.go#L169-L228) - ❌  Updates individual records, no rollback attempt, no documentation, returns modified records and error on failure
* [nanelo](https://github.com/libdns/nanelo/tree/cec9a30fee8c26143256016e82c9e3db8c727f35) - 🟨   Doesn't implement `RecordSetter`
* [domainnameshop](https://github.com/libdns/domainnameshop/blob/5200f129a0be27bd419e4df6ea677a983288c9fe/provider.go#L90-L118) - ❌  Updates individual records, no rollback attempt, no documentation, returns nil and error on failure
* [ionos](https://github.com/libdns/ionos/blob/8ffc7fb3619b419dda01b0dcbf18c9f00822224c/provider.go#L231-L249) - ❌  Updates individual records, no rollback attempt, no documentation, returns nil and error on failure. From a quick glance, this one can be fixed easily, since they have methods to batch update records, they're just not used here for whatever reason. I don't use this provider nor did I do any deep digging, so there are probably reasons they opted not to go this route.
* [ovh](https://github.com/libdns/ovh/blob/b6770e1a7959f837c7e2606192b7c42f64e34719/provider.go#L50-L68) - 🟨  Updates individual records, no rollback attempt, *has* documentation, returns nil and error on failure
* [scaleway](https://github.com/libdns/scaleway/blob/28d7237a7865ac1ea948682b4a0414f81f843879/provider.go#L48-L67) - ❌  Updates individual records, no rollback attempt, no documentation, returns modified records and error on failure
* [he](https://github.com/libdns/he/blob/932986f81a5d870169c95d038e3f0a71971cebd6/provider.go#L48-L62) - ❌  Updates individual records, no rollback attempt, no documentation, returns nil and error on failure
* [gcore](https://github.com/libdns/gcore/blob/9964da923833c06f4460a46512402754d5b96248/provider.go#L233-L273) - ❌  Updates individual records, no rollback attempt, no documentation, returns nil and error on failure
* [azure](https://github.com/libdns/azure/blob/721fd048b7f6db58820e18321b59578170e3c7a8/provider.go#L64-L78) - ❌  Updates individual records, no rollback attempt, no documentation, returns nil and error on failure
* [desec](https://github.com/libdns/desec/blob/3cac5c92acccd2af872b3c6e847d817a8e5c2745/provider.go#L200-L244) - ✅  Conforms!
* [infomaniak](https://github.com/libdns/infomaniak/blob/6eabf9cb56369df32694b04285548491a390df10/provider.go#L87-L153) - 🟨  Updates individual records, no rollback attempt, has documentation but it's confusing since it's the `RecordSetter` documentation but with some words changed and at a quick glance is easy to gloss over, returns nil and error on failure
* [duckdns](https://github.com/libdns/duckdns/blob/3a025cb0456370b4866d3b946c29036f4b049a00/provider.go#L70-L83) - ❌  Updates individual records, no rollback attempt, no documentation, returns nil and error on failure
* [alidns](https://github.com/libdns/alidns/blob/6b1428522560d0b4fdf07e7aa06939ed7af2caef/provider.go#L73-L99) - ❌  Updates individual records, no rollback attempt, no documentation, returns nil and error on failure

I skipped providers that haven't updated to 1.0.0(-beta) yet (since this behavior wasn't specified prior) and special cases like acmeproxy or acmedns.

I'm not trying to dunk on the authors of these implementations, I'd bet it's because these providers are missing bulk operation APIs. My point is that the current spec seems like it's not viable for implementers, causing the vast majority of them to be non-conformant. It's probably too late since 1.0 has stabilized, but it seems like `RecordSetter` should not make the promise of atomicity given the provider landscape.

mholt:
This introduces AtomicErr, and clarifies the documentation regarding atomicity. It is no longer expected/required for SetRecords, but encouraged, and AtomicErr can be returned to inform callers of the state of their zone.

### DIFF
--- a/libdns.go
+++ b/libdns.go
@@ -148,15 +148,6 @@ type RecordSetter interface {
 	// consistent state. Implementations should document their atomicity
 	// guarantees (or lack thereof).
 	//
-	// Calls to SetRecords are presumed to be atomic; that is, if err == nil,
-	// then all of the requested changes were made; if err != nil, then none of
-	// the requested changes were made, and the zone is as if the method was
-	// never called. Some provider APIs may not support atomic operations, so it
-	// is recommended that implementations synthesize atomicity by transparently
-	// rolling back changes on failure; if this is not possible, then it should
-	// be clearly documented that errors may result in partial changes to the
-	// zone.
-	//
 	// If SetRecords is used to add a CNAME record to a name with other existing
 	// non-DNSSEC records, implementations may either fail with an error, add
 	// the CNAME and leave the other records in place (in violation of the DNS
--- a/libdns.go
+++ b/libdns.go
@@ -140,12 +140,11 @@ type RecordSetter interface {
 	// err == nil, then all of the requested changes were made, and if err != nil,
 	// then the zone remains as if the method was never called. However, as very
 	// few providers offer batch/atomic operations, the actual result of a call
-	// where err != nil is undefined. If it's possible for implementations to
-	// synthesize atomicity by transparently rolling back changes on failure,
-	// this may not always succeed, but may be attempted to try to restore zone
-	// consistency. For calls that error atomically, implementations should return
-	// AtomicErr as the error so callers may know that their zone remains in a
-	// consistent state. Implementations should document their atomicity
+	// where err != nil is undefined. Implementations may implement synthetic
+	// atomicity that rolls back partial changes on failure ONLY if it can be
+	// done reliably. For calls that error atomically, implementations should
+	// return [AtomicErr] as the error so callers may know that their zone remains
+	// in a consistent state. Implementations should document their atomicity
 	// guarantees (or lack thereof).
 	//
 	// If SetRecords is used to add a CNAME record to a name with other existing
--- a/libdns.go
+++ b/libdns.go
@@ -136,6 +136,18 @@ type RecordSetter interface {
 	// should not blindly call SetRecords with the output of
 	// [libdns.RecordGetter.GetRecords].
 	//
+	// If possible, implementations should make SetRecords atomic, such that if
+	// err == nil, then all of the requested changes were made, and if err != nil,
+	// then the zone remains as if the method was never called. However, as very
+	// few providers offer batch/atomic operations, the actual result of a call
+	// where err != nil is undefined. If it's possible for implementations to
+	// synthesize atomicity by transparently rolling back changes on failure,
+	// this may not always succeed, but may be attempted to try to restore zone
+	// consistency. For calls that error atomically, implementations should return
+	// AtomicErr as the error so callers may know that their zone remains in a
+	// consistent state. Implementations should document their atomicity
+	// guarantees (or lack thereof).
+	//
 	// Calls to SetRecords are presumed to be atomic; that is, if err == nil,
 	// then all of the requested changes were made; if err != nil, then none of
 	// the requested changes were made, and the zone is as if the method was
@@ -290,3 +302,9 @@ func AbsoluteName(name, zone string) string {
 	}
 	return name + "." + zone
 }
+
+// AtomicErr should be returned as the error when a method errors
+// atomically. When this error type is returned, the caller can
+// know that their zone remains in a consistent state despite an
+// error.
+type AtomicErr error


worr:
Sure, this addresses my concerns. Personally, I'd drop the recommendation of synthesizing rollback, since it's error-prone especially for some providers, but I'm also not particularly bothered. Thanks a ton for addressing this, I appreciate it.

worr (approved):

mholt:
I will make the language a little more cautious about rollback then.

mholt:
One last thought... should it be called `AtomicError`? can't decide between the "Err" pattern I've seen in Go std lib, or just spelling things out because it's not like it's a frequently used identifier.

gucci-on-fleek (approved):

gucci-on-fleek:
> One last thought... should it be called `AtomicError`? can't decide between the "Err" pattern I've seen in Go std lib, or just spelling things out because it's not like it's a frequently used identifier.

I think that `AtomicErr` is probably better since it matches the stdlib, but `AtomicError` seems fine to me too.

mholt:
Sounds good, thanks both!

gjung56:
Is this pattern production ready?

I upgraded my libdns ovh package with SetRecords atomatic behavior, I used the AtomicErr type.
I used the head commit in my go.mod, until a tagged version is ready.

gucci-on-fleek:
@gjung56 

> Is this pattern production ready?
> 
> I upgraded my libdns ovh package with SetRecords atomatic behavior, I used the AtomicErr type. I used the head commit in my go.mod, until a tagged version is ready.

Ok, I've [tagged v1.1.0](https://github.com/libdns/libdns/releases/tag/v1.1.0) just now, so you should be able to use that instead. Hopefully @mholt doesn't mind :)

mholt:
Not at all, thank you for beating me to it!

PR #187, opened 2025-08-09
Clarify `SetRecords` behaviour

### LINKED ISSUE #186: Most implementations missing deletion behavior for SetRecords
According to the [RecordSetter.SetRecords](https://github.com/libdns/libdns/blob/6ab812eb60617ba3df584c0939987ad725f63424/libdns.go#L118) documentation, implementations are required to ensure that for any (name, type) pair in the input, those are the only records (of that kind) present in the resulting zone.

 > For any (name, type) pair in the input, SetRecords ensures that the only records in the output zone with that (name, type) pair are those that were provided in the input.
> In RFC 9499 terms, SetRecords appends, modifies, or deletes records in the zone so that for each RRset in the input, the records provided in the input are the only members of their RRset in the output zone.

If I'm not mistaken, Caddy (via certmagic) is the main consumer of this library. In its typical flow, when provisioning ECH, it usually calls SetRecords with a single record, and doesn't check beforehand (via GetRecords) for any existing records of the same name and type.

In the case of HTTPS with ECH, if someone tries to cover common scenarios for different versions of HTTP (2, 1.1, 3) with records addressing that, the implementation which follows the documentation (above cited) would make a mess.

For example, the [ovh](https://github.com/libdns/ovh/blob/5a73284aca79577f7a1c5866a434a79dbc0fc95c/client.go#L109) implementation takes care of deletion. Most other implementations don't.

This seems to work fine, so far, with most current provider implementations. I wonder if this discrepancy could lead to bugs or interoperability issues in less typical scenarios. Regarding this, what would be the guideline for future implementations?

gucci-on-fleek:
Some implementations implement `SetRecords` and `AppendRecords` identically, but this is incorrect. This commit updates the `SetRecords` documentation to clarify that `SetRecords` can only be implemented by a combination of `DeleteRecords` and `AppendRecords`, not `AppendRecords` alone.

Fixes #186.

### DIFF
--- a/libdns.go
+++ b/libdns.go
@@ -129,6 +129,25 @@ type RecordSetter interface {
 	// zone so that for each RRset in the input, the records provided in the input
 	// are the only members of their RRset in the output zone.
 	//
+	// SetRecords is distinct from [libdns.RecordAppender.AppendRecords] in that
+	// AppendRecords *only* adds records to the zone, while SetRecords may also
+	// delete records if necessary. Therefore, SetRecords behaves similarly to
+	// the following code:
+	//
+	//	func SetRecords(ctx context.Context, zone string, recs []Record) ([]Record, error) {
+	//		prevs, _ := p.GetRecords(ctx, zone)
+	//		toDelete := []Record{}
+	//		for _, prev := range prevs {
+	//			for _, new := range recs {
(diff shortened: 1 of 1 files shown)


mholt (approved):
Thanks! As usual, this is more thorough than I would have done myself, haha. But I think this is great.

Issue #178, opened 2025-06-07
`.toCAA()` doesn't handle CAA-values with spaces


NicolaiSoeborg:
For fancy CAA records, e.g. [ACME-CAA](https://xn--sb-lka.org/ACME-CAA/) the parsing logic in `.toCAA` doesn't handle "CAA-value" data that contains spaces (but quoted)

https://github.com/libdns/libdns/blob/6be57668e7bf10f6c31ab7ffcd7cc132766c1ee2/record.go#L157-L165

Example CAA record failing: `128 issue "letsencrypt.org; accounturi=..."`

gucci-on-fleek:
Hmm, good catch; thanks for letting us know!

PR #189, opened 2025-08-30
Add basic end-to-end tests and in-memory DNS provider for libdns

AndrianBdn:
I came across libdns while checking an outdated Caddy DNS provider module that depends on it.

I noticed there were no functional tests defined - only a recommendation to test. Because of that, many providers have issues.

To help, I added a small set of end-to-end tests and a simple in-memory (non-synchronized) DNS provider. It is a basic implementation that passes the tests, to test the tests. 

I also implemented these tests on my Cloudflare libdns provider fork:  
https://github.com/libdns/cloudflare/compare/master...AndrianBdn:cloudflare:master 

It turns out the Cloudflare DNS provider has some weird behavior due to unusual FQDN normalization in CNAME (this is actually very common problem - might we worth documenting if libdns clients expect dots in external CNAMEs or not). 

I've tried to test DigitalOcean implementation, however it quickly become evident that its current SetRecords implementation is just a hack: https://github.com/libdns/digitalocean/blob/master/provider.go#L64

I think the libdns community would benefit from having these basic e2e tests. The tests could also live in the template repo, but having a basic reference in-memory provider and test suite in the main library would likely be more useful.

I would open PRs with e2e tests and fixes to cloudflare, route53, maybe to digitalocean too once the framework is established.

### DIFF
--- a/libdnstest/README.md
+++ b/libdnstest/README.md
@@ -1,17 +1,32 @@
-# libdns End-to-End Testing
+# libdns Testing Framework
 
-This package provides reusable end-to-end testing utilities for libdns provider implementations.
+This package provides reusable testing utilities for libdns provider implementations.
 
 ## Usage
 
 ```go
-import "github.com/libdns/libdns/e2e"
+import "github.com/libdns/libdns/libdnstest"
 
-suite := e2e.NewTestSuite(provider, "example.com.")
+suite := libdnstest.NewTestSuite(provider, "example.com.")
 suite.RunTests(t)
 ```
 
-The TestSuite provides skip flags (`SkipMX`, `SkipSRV`, `SkipCAA`, `SkipNS`, `SkipSVCBHTTPS`) to exclude specific record types from testing.
+The TestSuite provides a `SkipRRTypes` map to exclude specific record types from testing:
+
+```go
+suite := libdnstest.NewTestSuite(provider, "example.com.")
+suite.SkipRRTypes = map[string]bool{
+    "MX":    true,  // Skip MX record tests
+    "SRV":   true,  // Skip SRV record tests
+    "CAA":   true,  // Skip CAA record tests
+    "NS":    true,  // Skip NS record tests
+    "SVCB":  true,  // Skip SVCB record tests
+    "HTTPS": true,  // Skip HTTPS record tests
+}
+suite.RunTests(t)
+```
+
+Note: Essential record types (A, CNAME, TXT) cannot be skipped as they are used by the testing framework itself.
 
 ## Providers Without ZoneLister
 
@@ -19,8 +34,8 @@ If your provider doesn't implement `ZoneLister`, use the `WrapNoZoneLister` help
 
 ```go
 provider := YourProvider{...} // implements RecordGetter, RecordAppender, RecordSetter, RecordDeleter
-wrappedProvider := e2e.WrapNoZoneLister(provider)
-suite := e2e.NewTestSuite(wrappedProvider, "example.com.")
+wrappedProvider := libdnstest.WrapNoZoneLister(provider)
+suite := libdnstest.NewTestSuite(wrappedProvider, "example.com.")
 suite.RunTests(t) // ListZones test will be skipped automatically
 ```
 
@@ -37,7 +52,7 @@ type MyRecord struct {
 func (r MyRecord) RR() libdns.RR { return r.RR }
 
 // Configure custom record constructor
-suite := e2e.NewTestSuite(provider, "example.com.")
+suite := libdnstest.NewTestSuite(provider, "example.com.")
 suite.AppendRecordFunc = func(record libdns.Record) libdns.Record {
     return MyRecord{
         RR:    record.RR(),
@@ -72,13 +87,13 @@ The test suite automatically cleans up test records using `AttemptZoneCleanup()`
 
 **Use dedicated test zones when working with real DNS providers.**
 
-## Dummy Provider
+## Example Provider
 
 ```go
-import "github.com/libdns/libdns/e2e/dummy"
+import "github.com/libdns/libdns/libdnstest/example"
 
-provider := dummy.New("example.com.")
+provider := example.New("example.com.")
 records, err := provider.GetRecords(ctx, "example.com.")
 ```
 
-The dummy provider implements all libdns interfaces using in-memory storage. It serves as a double-entry system to ensure there is some implementation that can pass these tests. The dummy provider does not guarantee DNS compliance, but works for the currently defined tests.
+The example provider implements all libdns interfaces using in-memory storage. It serves as a double-entry system to ensure there is some implementation that can pass these tests. The example provider does not guarantee DNS compliance, but works for the currently defined tests.
--- a/libdnstest/example/example_test.go
+++ b/libdnstest/example/example_test.go
@@ -1,23 +1,23 @@
-package dummy_test
+package example_test
 
 import (
 	"context"
 	"testing"
 
 	"github.com/libdns/libdns"
-	"github.com/libdns/libdns/e2e"
-	"github.com/libdns/libdns/e2e/dummy"
+	"github.com/libdns/libdns/libdnstest"
+	"github.com/libdns/libdns/libdnstest/example"
 )
 
-func TestDummyProvider(t *testing.T) {
-	provider := dummy.New("example.com.")
-	testSuite := e2e.NewTestSuite(provider, "example.com.")
+func TestExampleProvider(t *testing.T) {
+	provider := example.New("example.com.")
+	testSuite := libdnstest.NewTestSuite(provider, "example.com.")
 	testSuite.RunTests(t)
 }
 
-func TestDummyProviderWithSkippedRecords(t *testing.T) {
-	provider := dummy.New("example.com.")
-	testSuite := e2e.NewTestSuite(provider, "example.com.")
+func TestExampleProviderWithSkippedRecords(t *testing.T) {
+	provider := example.New("example.com.")
+	testSuite := libdnstest.NewTestSuite(provider, "example.com.")
 
 	// Example: skip some types that some provider may not implement
 	testSuite.SkipRRTypes = map[string]bool{
@@ -34,15 +34,15 @@ func TestDummyProviderWithSkippedRecords(t *testing.T) {
 }
 
 type noZoneListerProvider struct {
-	provider *dummy.Provider
+	provider *example.Provider
 }
 
-func TestDummyProviderNoZoneLister(t *testing.T) {
-	// let's pretent we have provider that does not implement ZoneListener
-	provider := &noZoneListerProvider{provider: dummy.New("example.com.")}
+func TestExampleProviderNoZoneLister(t *testing.T) {
+	// let's pretend we have provider that does not implement ZoneLister
+	provider := &noZoneListerProvider{provider: example.New("example.com.")}
 
-	// this how how we test it:
-	suite := e2e.NewTestSuite(e2e.WrapNoZoneLister(provider), "example.com.")
+	// this is how we test it:
+	suite := libdnstest.NewTestSuite(libdnstest.WrapNoZoneLister(provider), "example.com.")
 	suite.RunTests(t)
 }
 
--- a/libdnstest/example/provider.go
+++ b/libdnstest/example/provider.go
@@ -1,5 +1,5 @@
-// Package dummy provides an in-memory implementation of all libdns interfaces for testing.
-package dummy
+// Package example provides an in-memory implementation of all libdns interfaces for testing.
+package example
 
 import (
 	"context"
--- a/libdnstest/libdnstest.go
+++ b/libdnstest/libdnstest.go
@@ -1,17 +1,17 @@
-// Package e2e provides end-to-end testing utilities for libdns provider implementations.
+// Package libdnstest provides testing utilities for libdns provider implementations.
 //
 // These tests create, modify, and delete DNS records with names like "test-append",
 // "test-set", "test-delete", and "test-lifecycle" to validate provider behavior.
 //
 // For real DNS provider implementations, use dedicated test zones since the tests
-// will modify DNS records. The dummy provider uses in-memory storage and is completely safe.
+// will modify DNS records. The example provider uses in-memory storage and is completely safe.
 //
 // Tests run sequentially (not in parallel) to avoid conflicts when testing real
 // DNS providers that interact with external services.
 //
 // # Usage
 //
-//	suite := e2e.NewTestSuite(yourProvider, "example.com.")
+//	suite := libdnstest.NewTestSuite(yourProvider, "example.com.")
 //	suite.RunTests(t)
 //
 // # Provider Without ZoneLister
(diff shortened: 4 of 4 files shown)


gucci-on-fleek:
> I noticed there were no functional tests defined - only a recommendation to test. Because of that, many providers have issues.
>
> To help, I added a small set of end-to-end tests and a simple in-memory (non-synchronized) DNS provider. It is a basic implementation that passes the tests, to test the tests.

Nice, thanks! My tests for the `rfc2136` provider ([`conversion_test.go`](https://github.com/libdns/rfc2136/blob/master/conversion_test.go), [`provider_test.go`](https://github.com/libdns/rfc2136/blob/master/provider_test.go)) look a little bit more comprehensive at the moment, so you might want to copy some of that over to your PR.

> I think the libdns community would benefit from having these basic e2e tests. [...] having a [...] test suite in the main library would likely be more useful.

I agree that this is the best place to store the tests, because if we find common bugs, we can just update this single repo instead of having to make PRs to every repo to add new tests.

> The tests could also live in the template repo

It might be a good idea to add a GitHub Actions workflow to the template repo so that new users are pushed to implement e2e tests (or make the conscious decision to delete the file). [The required configuration is _really_ short for this.](https://github.com/libdns/rfc2136/blob/master/.github/workflows/go.yml)

> basic reference in-memory provider

Hmm, I'm unsure if this is a good idea. DNS is essentially 100 different edge cases piled on top of each other, so it will be fairly tricky to make a minimal in-memory  provider that will work like a “real” DNS provider. My suggestion would be to (either)

1. Use the [`rfc2136` provider](https://github.com/libdns/rfc2136) for this purpose. There's almost the same number of lines of code for `dummy` (124) and `rfc2136` (224), but `rfc2136` uses a real DNS server, so it should properly validate all the weird edge cases. Feel free to email me if you want API keys to test against my server—I've already given access to `libdns/rfc2136` and DNSControl for similar purposes.

2. Implement a basic Bind zonefile provider. This has the advantage of being intrinsically useful, and we could shell out to [`named-checkzone`](https://man.archlinux.org/man/named-checkzone.1) during the tests to validate with real DNS software (that runs completely locally).

But if you think that the in-memory provider is the best solution, then I'm sure that I could be convinced.

> I would open PRs with e2e tests and fixes to cloudflare, route53, maybe to digitalocean too once the framework is established.

:+1:

e2e/e2e.go
```diff
@@ -0,0 +1,506 @@
...
+)
+
+// RecordProvider represents a basic libdns provider that handles DNS records.
+// This is the most common type of provider implementation.
+type RecordProvider interface {
+	libdns.RecordGetter
+	libdns.RecordAppender
+	libdns.RecordSetter
+	libdns.RecordDeleter
+}
```
gucci-on-fleek, line 62: I've tested this with `libdns/rfc2136`, and I get the following error

```log
# github.com/libdns/rfc2136/e2e [github.com/libdns/rfc2136/e2e.test]
./rfc2136_test.go:20:34: cannot use provider (variable of struct type rfc2136.Provider) as e2e.RecordProvider value in argument to e2e.NewRecordTestSuite: rfc2136.Provider does not implement e2e.RecordProvider (method AppendRecords has pointer receiver)
FAIL    github.com/libdns/rfc2136/e2e [build failed]
FAIL
```

[because `libdns/rfc2136` uses](https://github.com/libdns/rfc2136/blob/d49ba0aa4bf79391aaacf3ec1db5c6bdafff5edf/provider.go#L119)

```go
func (p *Provider) AppendRecords(ctx context.Context, zone string, records []libdns.Record) ([]libdns.Record, error) {
	...
}
```

I don't really know much about Go, so I have no idea if `(p *Provider) AppendRecords` is actually correct or not.

gucci-on-fleek (commented):
Ok, [I've added some code to my fork of `libdns/rfc2136`](https://github.com/gucci-on-fleek/libdns-rfc2136/commit/a5151f7e3b30fb2a1fdb3805a6315d8d83df37ff), and the [tests all ran successfully](https://github.com/gucci-on-fleek/libdns-rfc2136/actions/runs/17354032687/job/49264397491). There are still a few things that I'm unsure about, but this definitely appears to be on the right track. Thanks again!

e2e/e2e.go
gucci-on-fleek: Ugh, I _knew_ that I was missing something obvious; thanks!

e2e/e2e.go
```diff
@@ -0,0 +1,506 @@
...
+
+	for _, current := range currentTestSetRecords {
+		currentRR := current.RR()
+		if currentRR.Data == "192.0.2.1" || currentRR.Data == "192.0.2.2" {
+			t.Errorf("Old record data still exists: %s", currentRR.Data)
+		}
+		if currentRR.Data != "192.0.2.3" {
+			t.Errorf("Expected updated record data 192.0.2.3, got %s", currentRR.Data)
+		}
+	}
```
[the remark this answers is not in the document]
AndrianBdn: Not sure I got what you have meant by parametrization. But it is a good point. I've referenced IPs from the structs, to avoid copy-pasting them.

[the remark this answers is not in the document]
gucci-on-fleek: > I've referenced IPs from the structs, to avoid copy-pasting them.

Perfect, thanks!

AndrianBdn:
Thanks for the comment @gucci-on-fleek and for looking at this PR.

We're already aligned on test placement. I'm fine with a GitHub Actions template too - most providers will still need some zones & secrets set up by maintainers, so CI isn't the hardest part here.

> [About the in-memory provider, DNS complexity, etc]

I called these tests "e2e", but they're really just sanity checks. They don't do real DNS queries - we just CRUD records via libdns interfaces and check consistency. I think that's still useful for the libdns community, since many providers are incomplete or have quirks. Maybe we should rename them to 'sanity'? )

I don't see much value in testing actual DNS behavior. That's the DNS server/service job (AWS, GCP, Azure, etc.). And even if we find oddities (TTL rounding, delayed propagation, partial support of certain values, etc.), getting them fixed is slow, and most would just end up as test config flags (TestShouldIgnoreTTL, TestSleepAfterAdd: 4s).

On the "test provider”: I like the idea of keeping libdns self-contained. You clone, run tests, and it just works. Adding more cases is straightforward. Depending on an external provider like rfc2136, which itself depends on a live server, adds complexity - and even if that server is spec-perfect, most big DNS providers aren't.

Open to input: should we keep it small and focused, or go all-in with more record types and real DNS checks?

gucci-on-fleek:
@AndrianBdn 

> > [About the in-memory provider, DNS complexity, etc]
> 
> I called these tests "e2e", but they're really just sanity checks. They don't do real DNS queries - we just CRUD records via libdns interfaces and check consistency. I think that's still useful for the libdns community, since many providers are incomplete or have quirks. Maybe we should rename them to 'sanity'? )
> 
> […]
> 
> On the "test provider”: I like the idea of keeping libdns self-contained. You clone, run tests, and it just works. Adding more cases is straightforward. Depending on an external provider like rfc2136, which itself depends on a live server, adds complexity - and even if that server is spec-perfect, most big DNS providers aren't.

I guess my question is, what's the point of the in-memory provider? If the only purpose of it is to make adding new tests easier, it doesn't really seem worth it—we'd want to test any new tests with at least 1 real provider before adding it upstream (here), so having the in-memory provider just adds another provider that can potentially fail.

> I don't see much value in testing actual DNS behavior. That's the DNS server/service job (AWS, GCP, Azure, etc.). And even if we find oddities (TTL rounding, delayed propagation, partial support of certain values, etc.), getting them fixed is slow, and most would just end up as test config flags (TestShouldIgnoreTTL, TestSleepAfterAdd: 4s).

By testing actual DNS behaviour, I meant more something like that an actual DNS server will block us from setting a `CNAME` record for the same name as a `A`/`TXT` record, won't let us set `ipv4hint=2001:db8::1`, won't let us set an `MX` record with an IP address for a target, etc.; not making sure that the providers aren't lying to us. Using any real DNS provider should stop us from adding such test cases, but these will all work with the current in-memory provider (I think?).

> Open to input: should we keep it small and focused, or go all-in with more record types and real DNS checks?

I think that we will probably want tests at least as comprehensive as the [`rfc2136`](https://github.com/libdns/rfc2136) ones ([`conversion_test.go`](https://github.com/libdns/rfc2136/blob/master/conversion_test.go), [`provider_test.go`](https://github.com/libdns/rfc2136/blob/master/provider_test.go)). I think that the minimum required to get this merged is probably:

1. Test to make sure that implementations are handling `SetRecords` vs `AppendRecords` correctly (#186). I think that you've already done this :ballot_box_with_check:

2. Test every single `Record` type and test setting every single field. This should be trivial to implement, just lots of annoying copy-and-pasting. You can probably [copy this list](https://github.com/libdns/rfc2136/blob/d49ba0aa4bf79391aaacf3ec1db5c6bdafff5edf/provider_test.go#L128-L206), or just copy-and-paste some of your current records and then change their types.

Eventually (for a later PR if someone is interested), it would probably be a good idea to do the following:

1. Test weird edge cases, like [this list from DNSControl](https://github.com/StackExchange/dnscontrol/blob/main/integrationTest/integration_test.go).
2. Add tests for previously-fixed issues (like #178) to prevent any future regressions.

e2e/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Test Coverage
+
+- **ListZones**: Lists available zones (FullProvider only)
+- **GetRecords**: Retrieves records from a zone  
+- **AppendRecords**: Creates new records (uses `test-append*` names)
+- **SetRecords**: Creates/updates/deletes records by (name,type), preserves unrelated records (uses `test-set*` names)
+- **DeleteRecords**: Creates then deletes records (uses `test-delete*` names) 
+
+For **real DNS providers**, use dedicated test zones since tests create/modify/delete DNS records.
```
gucci-on-fleek, line 57: I think that we should use stronger language here (feel free to reword however you want)

```suggestion
> [!WARNING]
> Running these tests will **delete all records** currently in the zone, so you should **only** run the tests on dedicated test zones.
```

Even if only `test-*` records are modified, I'd still prefer to warn against all records in case we ever change something (or add a bug).

e2e/README.md
```diff
@@ -0,0 +1,84 @@
...
+suite.RunRecordTests(t)
+```
+
+## Test Coverage
+
+- **ListZones**: Lists available zones (FullProvider only)
+- **GetRecords**: Retrieves records from a zone  
+- **AppendRecords**: Creates new records (uses `test-append*` names)
+- **SetRecords**: Creates/updates/deletes records by (name,type), preserves unrelated records (uses `test-set*` names)
+- **DeleteRecords**: Creates then deletes records (uses `test-delete*` names)
```
gucci-on-fleek, line 55: I'd slightly prefer using a description list here. There's no dedicated Markdown syntax for this, but embedding the HTML in Markdown works fine:

```markdown
<dl>
<dt>head 1
<dd>description 1
<dt>head 2
<dd>description 2
</dl>
```

But this is a _weird_ preference, so feel free to ignore :)

e2e/e2e.go
```diff
@@ -0,0 +1,506 @@
...
+func (ts *TestSuite) cleanupRecords(t *testing.T, ctx context.Context, records []libdns.Record) {
+	if len(records) == 0 {
+		return
+	}
+
+	_, err := ts.recordProvider.DeleteRecords(ctx, ts.Zone, records)
+	if err != nil {
+		t.Logf("Warning: cleanup failed: %v", err)
+	}
+}
```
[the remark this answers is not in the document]
gucci-on-fleek: Hmm, I'm still not sure if this is the best approach. If the program crashes during testing, then any records won't be cleaned up, which could interfere with following runs, leading to flaky tests.

e2e/README.md
AndrianBdn: Thanks, updated formatting

gucci-on-fleek:
@AndrianBdn 

> thanks for you suggestions!
> 
>  * Adjusted cleanup logic.
> 
>  * Added support for more DNS record types. I had to put in exclusions. Better to let providers adapt tests and keep other records support as an open issue.

Thanks, I'll look at the code a little while later.

>  * After thinking it through, I’m going to support only one Provider interface set. Incomplete providers can work around this, or just add the missing method.

I think that it might be easiest to just leave out `ListZones` entirely. Only [8](https://github.com/search?q=org%3Alibdns+%2Ffunc.*ListZones%2F+NOT+path%3Atest&type=code) out of [79](https://github.com/orgs/libdns/repositories?type=all) providers currently support it, and it's not used by CertMagic/Caddy for anything.

> > I meant more something like that an actual DNS server will block us from setting a CNAME record for the same name as a A/TXT record, won't let us set ipv4hint=2001:db8::1, won't let us set an MX record with an IP address for a target, etc.; not making sure that the providers aren't lying to us.
> 
> I don’t think that’s really a libdns issue. If a service allows something invalid (like IP in MX), that’s more about the service than libdns. In theory some validation can live in libdns — e.g. MX or SVCB record checks.
> 
> I’m not attached to the in-memory dummy provider. The only benefit is that it creates a double-entry system for tests and a dummy implementation. But since I haven’t tested with any real provider that supports all record types, the dummy provider may not be correct indeed.

Ok, I still don't really see the point of the in-memory provider, but you've already written the code, and it should be trivial to maintain, so I'm okay with including it.

> I can remove it, and we can add a note that e2e tests should be verified with [rfc2136](https://github.com/libdns/rfc2136) as a reference implementation of a libdns provider (can libdns maintainers give it such status?). 

I'm not really sure that we need a reference provider—in theory `libdns` should behave the same everywhere, so if 2 providers disagree, then there's probably a bug somewhere. And different people are more comfortable testing with different providers, so it's probably easier to let people use whatever they prefer.

> BTW it would also help to have a Dockerfile + script there to start a DNS server (bind?) with a preconfigured zone and key.

Hmm, that's definitely a good idea, I'll look into that.

AndrianBdn:
> I'm not really sure that we need a reference provider—in theory libdns should behave the same everywhere, so if 2 providers disagree, then there's probably a bug somewhere.

Maybe I am over-thinking it, but here is my concern:

When someone submits a PR that changes tests, how do we know the change is valid? One option is to require them to show it works against some provider. A reference provider gives us a common baseline. It could be in-memory (although it is limited in DNS sense and probably buggy), or rfc2136, or something else. This is not about "truth" in DNS - only about having something consistent to measure test changes against. 

Obviously we can introduce a policy, that requires to adopt new tests in any two providers. But that feels like unnecessary friction.

mholt:
Sorry I've been absent from this for a while!

> I think that it might be easiest to just leave out ListZones entirely.

Could it be optional? Maybe with a warning or something if it's not implemented, as opposed to a hard fail?

> When someone submits a PR that changes tests, how do we know the change is valid? One option is to require them to show it works against some provider. A reference provider gives us a common baseline. It could be in-memory (although it is limited in DNS sense and probably buggy), or rfc2136, or something else. This is not about "truth" in DNS - only about having something consistent to measure test changes against.

This is a good point IMO.

Overall I'm in favor of this change, although I might have a nit about the folder structure/placement. I dunno -- will think on it.

AndrianBdn:
@mholt @gucci-on-fleek I introduced a different way to skip ZoneLister test

gucci-on-fleek:
> One question: should providers expect the trailing dot in FQDNs (e.g., CNAME targets)?

Yes, see https://docs.dnscontrol.org/language-reference/why-the-dot for a better explanation.

> I haven’t found anything in libdns about it. 

Hmm, good point; I'll open a PR for that shortly.

> For our CNAME records in tests, do we keep the trailing dot or drop it?

Definitely keep it.

PR #190, opened 2025-09-01
Make `Record.RR()` always deterministic

gucci-on-fleek:
`SvcParams.String()` used to iterate over keys in a random order, which meant that calling `ServiceBinding.RR()` could return different `Data` for identical records. This commit sorts the keys in lexicographic order, making the result deterministic.

Furthermore, since it seems useful for code to assume that it can safely roundtrip between `Record.RR()` and `RR.Parse()` and that `Record.RR()` is deterministic, I've now documented this. These are both already the case, but a formal guarantee should make external usage easier.

Related to #189.

---

[RFC 9460 §2.1](https://datatracker.ietf.org/doc/html/rfc9460#section-2.1-11) says that we're allowed to place the keys in any order that we want, so I decided to sort the keys lexicographically since that was the easiest to implement. However, [the canonical order is to sort by the key index](https://datatracker.ietf.org/doc/html/rfc9460#section-2.2-5), but this would require for us to hardcode [the key↔index mapping](https://www.iana.org/assignments/dns-svcb/dns-svcb.xhtml).

### DIFF
--- a/record.go
+++ b/record.go
@@ -2,7 +2,9 @@ package libdns
 
 import (
 	"fmt"
+	"maps"
 	"net/netip"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


mholt (approved):
This is a good change. Thank you!

PR #193, opened 2025-09-12
Fix cleanup of SRV and SVCB records in libdnstest

AndrianBdn:
Fixes a cleanup bug in SRV and SVCB records in libdnstest, introduced in https://github.com/libdns/libdns/pull/189.

The bug slipped through because tests didn’t check for unknown leftovers, and the in-memory provider had no persistence to expose it.

The fix corrects cleanup logic and adds an optional strict mode: zones must be empty after operations. This is good for the in-memory provider. It’s not defaulted, since not everyone can dedicate a separate test zone (they cost about $6/year on AWS).


gucci-on-fleek:
[The tests all pass with rfc2136](https://github.com/gucci-on-fleek/libdns-rfc2136/actions/runs/17689780824/job/50281195565), so LGTM. Thanks again!

gucci-on-fleek (approved):

mholt (approved):

PR #194, opened 2025-09-13
Another clarification about trailing dots

AndrianBdn:
I hit a bug in a third-party provider caused by a missing trailing dot in the zone name. The zone was read from the environment along with the secret token, and without the dot everything broke. To make this obvious, I added a check so tests always expect dot-terminated zone names.

That leads to the real question about ZoneLister: do we require dot termination, or do we treat zone names as arbitrary strings? If a zone name can be something like a UUID, then checking for dots is meaningless, and the only real requirement is "non-empty". 

The tests today assume zone name is FQDN with dot termination, but if that's not the contract, I'll relax the check.

AbsoluteName is another inconsistency. It doesn't check for trailing dots at all. Options are: enforce with error, silently append, or leave it to the caller (current way). I haven't touched it, but my preference is to enforce consistency one way or the other.

### DIFF
--- a/libdns.go
+++ b/libdns.go
@@ -261,6 +261,7 @@ type ZoneLister interface {
 
 // [Zone] is a generalized representation of a DNS zone.
 type Zone struct {
+	// Name is the fully qualified domain name of the zone, including the trailing dot (e.g., "example.com.")
 	Name string
 }
 
--- a/libdns.go
+++ b/libdns.go
@@ -303,6 +303,11 @@ func AbsoluteName(name, zone string) string {
 	if zone == "" {
 		return strings.Trim(name, ".")
 	}
+	if !strings.HasSuffix(zone, ".") {
+		// "zone" should always be a FQDN (with a trailing dot), but we don't
+		// enforce this anywhere, so we'll go ahead and correct it if needed.
+		zone += "."
+	}
 	if name == "" || name == "@" {
 		return zone
 	}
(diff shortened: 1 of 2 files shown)


gucci-on-fleek:
> That leads to the real question about ZoneLister: do we require dot termination,

Without the trailing dot is fine-*ish* here since no one would ever confuse this with a relative name, but it's still always better to include the trailing dot, so this is a :+1: from me.

> or do we treat zone names as arbitrary strings? If a zone name can be something like a UUID, then checking for dots is meaningless, and the only real requirement is "non-empty".

Hmm, that's a good point. I'm pretty sure that we want to enforce that the zone name is an actual domain name, but if someone has a good argument for why we should also allow UUIDs, I don't think that I'd be opposed to that.

> AbsoluteName is another inconsistency. It doesn't check for trailing dots at all. Options are: enforce with error, silently append, or leave it to the caller (current way). I haven't touched it, but my preference is to enforce consistency one way or the other.

We should definitely make sure that `AbsoluteName` always returns something with the trailing dot, since otherwise `AbsoluteName(AbsoluteName(s)) != AbsoluteName(s)`. It looks like this is a regression caused by https://github.com/libdns/libdns/pull/153/commits/aba375c6ca567a175a2fb31548a225d1b1cc1887 for the case where the zone name doesn't end in a trailing dot.

gucci-on-fleek (approved):
I've gone ahead and added a fix for `AbsoluteName()`, but this is your PR, so please feel free to modify it if it doesn't look quite right to you.

Otherwise, LGTM, and thanks again.

PR #199, opened 2025-11-09
add allinkl provider

ljelschen:
this is the implementation for https://all-inkl.com


provider.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/libdns/libdns"
+)
+
+// TODO: Providers must not require additional provisioning steps by the callers; it
+// should work simply by populating a struct and calling methods on it. If your DNS
+// service requires long-lived state or some extra provisioning step, do it implicitly
+// when methods are called; sync.Once can help with this, and/or you can use a
+// sync.(RW)Mutex in your Provider struct to synchronize implicit provisioning.
+
+// Provider facilitates DNS record manipulation with <TODO: PROVIDER NAME>.
```
gucci-on-fleek:
```suggestion
// Provider facilitates DNS record manipulation with allinkl.
```

gucci-on-fleek (changes requested):
Looks like a good start; thanks for the contribution! I've made a few specific comments below, but in general, make sure to follow the advice of the Go comments in the template code.

Issue #200, opened 2025-11-18
make generic handling of records possible


mjl-:
Hi!

I've updated github.com/mjl-/dnsclay to use the libdns v1 API. It is mostly great, but I've run into a problem: The `RR.Parse` method can fail for some records that I wouldn't consider problematic (at least for generic DNS handling). This is because some of the `to<Type>` methods, like `RR.toSRV()`, try to parse the name/label of the record, and fail if it isn't in an expected form (i.e. it expects something like `_<service>._<transport>.<name>`, and if the two first sections are missing, the parsing fails). Actual DNS zones can have a SRV record for any  DNS name, not necessarily in the form expected. Providers (like rfc2136) are calling `RR.Parse()` for `GetRecords`, and propagate the error, so the `GetRecords` call may be failing entirely at retrieving records from a zone.

Related to that, some `to<Type>` methods (for SRV/SVCB/HTTPS) are modifying the DNS name of the record based on the value. This is a problem for generic handling, where the record is fetched from the source (potentially changing the name), and then deleted/updated (where the DNS name to be updated doesn't actually exist, because it was changed).

It's probably not a great idea for compatibility to change the current record name mangling for srv/svcb, given that the behaviour is in v1. I think it would be desirable for libdns if `RR.Parse` wouldn't fail on records that are in actual DNS zones (like the SRV record mentioned). Though perhaps that is not possible given the promises in the API.

Finally, it would help dnsclay with generic DNS record handling if there was an option to just have `RR.Parse` leave the record as is, like this: https://github.com/mjl-/libdns/blob/master/record.go#L129. Providers calling `RR.Parse` would just work as intended. Instead of `if true`, it could be based on a global variable, or possibly hidden and with a global `Set`-method to change it. I could make a PR for that. My current approach of a libdns-fork, with a replace-statement in the go.mod of dnsclay is problematic due to `go install` not accepting such replace directives and failing those compilations. Would adding a variable for a "raw" `RR.Parse` method be an option? Other ideas are welcome as too.

Tagging @mholt and @gucci-on-fleek because of earlier work on this.

mholt:
Since #210 has been merged, is this issue basically resolved? I think we're waiting for a real-world example where the SRV issue is a real problem, especially now that RR.Error exists, and the unparsed RR can be used.

Will close this, but we can reopen if I misunderstood.

PR #210, opened 2026-01-15
Never return an error in `RR.Parse()`

gucci-on-fleek:
This sounded like a good idea when we discussed this, but after writing out the code, I'm not so sure any more, since just throwing out the errors looks kinda ugly.

Maybe an alternative solution would be to add a special `ParseError` struct?

```go
// Parse returns a type-specific structure for this RR, if it is
// a known/supported type. Otherwise, it returns itself.
//
// Callers will typically want to type-assert (or use a type switch on)
// the return value to extract values or manipulate it.
//
// In the event of a parsing error, a special [ParseError] struct is returned,
// without indicating an error. This means that “err” will always be nil (the
// current signature is maintained for backwards compatibility).
func (r RR) Parse() (record Record, err error) {
	switch r.Type {
	case "A", "AAAA":
		record, err = r.toAddress()
	case "CAA":
		record, err = r.toCAA()
	case "CNAME":
		record, err = r.toCNAME()
	case "HTTPS", "SVCB":
		record, err = r.toServiceBinding()
	case "MX":
		record, err = r.toMX()
	case "NS":
		record, err = r.toNS()
	case "SRV":
		record, err = r.toSRV()
	case "TXT":
		record, err = r.toTXT()
	default:
		record = r
	}

	if err != nil {
		record, err = r.toParseError(err)
	}
	return
}

type ParseError struct {
	Name  string
	Type  string
	TTL   time.Duration
	Data  string
	Error error
}

func (c ParseError) RR() RR {
	return RR{
		Name: c.Name,
		TTL:  c.TTL,
		Type: c.Type,
		Data: c.Data,
	}
}

func (r RR) toParseError(err error) ParseError {
	return ParseError{
		Name:  r.Name,
		Type:  r.Type,
		TTL:   r.TTL,
		Data:  r.Data,
		Error: err,
	}
}
```

Seems kinda weird to define a specific error struct, but this way we aren't hiding the error message, but naïve implementations can freely ignore them and just treat them as opaque `RR`s.

See also #200.

### DIFF
--- a/record.go
+++ b/record.go
@@ -111,6 +111,10 @@ type RR struct {
 	// Implementations are not expected to support RFC 3597 “\#” escape
 	// sequences, but may choose to do so if they wish.
 	Data string `json:"data"`
+
+	// If parsing the RR into a more specific type failed, this field will
+	// contain the error. If parsing was successful, this field will be nil.
+	Error error `json:"error,omitempty"`
 }
 
 // RR returns itself. This may be the case when trying to parse an RR type
@@ -122,27 +126,39 @@ func (r RR) RR() RR { return r }
 //
 // Callers will typically want to type-assert (or use a type switch on)
 // the return value to extract values or manipulate it.
-func (r RR) Parse() (Record, error) {
+//
+// In the event of a parsing error, the [RR] will be returned as-is (unparsed),
+// with the error stored in the [RR.Error] field. This means that “err” will
+// always be nil (the current signature is maintained for backwards
+// compatibility).
+func (r RR) Parse() (record Record, err error) {
 	switch r.Type {
 	case "A", "AAAA":
-		return r.toAddress()
+		record, err = r.toAddress()
 	case "CAA":
-		return r.toCAA()
+		record, err = r.toCAA()
 	case "CNAME":
-		return r.toCNAME()
+		record, err = r.toCNAME()
 	case "HTTPS", "SVCB":
-		return r.toServiceBinding()
+		record, err = r.toServiceBinding()
 	case "MX":
-		return r.toMX()
+		record, err = r.toMX()
 	case "NS":
-		return r.toNS()
+		record, err = r.toNS()
 	case "SRV":
-		return r.toSRV()
+		record, err = r.toSRV()
 	case "TXT":
-		return r.toTXT()
+		record, err = r.toTXT()
 	default:
-		return r, nil
+		record = r
 	}
+
+	if err != nil {
+		r.Error = err
+		record, err = r, nil
+	}
+
+	return
 }
 
 func (r RR) toAddress() (Address, error) {
--- a/record_test.go
+++ b/record_test.go
@@ -674,3 +674,51 @@ func TestRRDataZeroValues(t *testing.T) {
 		}
 	}
 }
+
+func TestRRParseErrorField(t *testing.T) {
+	for _, test := range []RR{
+		{
+			Name: "example.com",
+			TTL:  1 * time.Hour,
+			Type: "A",
+			Data: "invalid-ip-address",
+		},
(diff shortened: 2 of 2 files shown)


mholt:
I envision user code being something like:

```go
for _, rec := range records {
	switch r := rec.(type) {
	case libdns.Address:
	case libdns.CNAME:
	case libdns.MX:
	case libdns.RR:
		return r.Error // likely parse error
	}
}
```

I guess there could still be an `if r.Error != nil` check rather than just assuming the `RR` means there was an error, but given our current API contract where we discourage returning `RR` values, I dunno, maybe this is fine. What do you think?

gucci-on-fleek:
@mholt 

> I guess there could still be an `if r.Error != nil` check rather than just assuming the `RR` means there was an error, but given our current API contract where we discourage returning `RR` values, I dunno, maybe this is fine. What do you think?

`RR` values should still be somewhat common, since there are [lots of other RRtypes](https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4) that we don't define. Here's the list (with counts) of the various RRtypes that I'm currently using for `maxchernoff.ca`

```log
    191 RRSIG
     63 NSEC
     58 TLSA
     29 TXT
     15 MX
     15 AAAA
     15 A
     14 HTTPS
      5 NS
      4 SRV
      3 TSIG
      3 CNAME
      3 CAA
      2 SOA
      1 SVCB
      1 SSHFP
      1 LOC
      1 DNSKEY
      1 CDS
      1 CDNSKEY
      1 ZONEMD
```

To be fair, I'm using lots of obscure RRtypes for the sole purpose of using lots of obscure RRtypes, so this probably isn't a common scenario :). But I think that using other RRtypes is sufficiently common that we probably shouldn't recommend blindly using `case libdns.RR: return r.Error`.

For the broader `RR.Error` versus `ParseError` question, I'm not really sure? The new `ParseError` struct option seems like a _slightly_ better option to me, but you're vastly more familiar with what idiomatic Go code looks like than I am, so I'll defer to you here.

mholt:
That's a good point, so I guess the `if r.Error != nil` check would still be required.

My main concern with `ParseError` is having a different type for every kind of error. (I'm not sure what other types there would be, but, you never know).

Is the Error check so bad? :sweat_smile: I think it's still my favorite in terms of flexibility, but I admit the elegance of a `ParseError` struct is tempting.

gucci-on-fleek:
@mholt 

> My main concern with `ParseError` is having a different type for every kind of error. (I'm not sure what other types there would be, but, you never know).

We could always call it an `ErrorRR` (or `RRError`).

> Is the Error check so bad? 😅 I think it's still my favorite in terms of flexibility, but I admit the elegance of a `ParseError` struct is tempting.

Sure, I'm fine with `RR.Error`. Probably the biggest advantage of that option is that it makes it easier to ignore errors; the biggest disadvantage is that it also makes it easier to ignore errors.

gucci-on-fleek:
Ok, I've pushed a new version with `RR.Error`. I wrote the runtime code myself, but I let Copilot handle writing the tests (although I still reviewed the code myself).

mholt (approved):
> Sure, I'm fine with RR.Error. Probably the biggest advantage of that option is that it makes it easier to ignore errors; the biggest disadvantage is that it also makes it easier to ignore errors.

Ha -- true, although, you could also just as well ignore an `error` return value anywhere else. An `RRError` struct is tempting, but then again you also have to add it to your type-assertion, which if you didn't, would _also_ be ignored.

And the nice thing about the `RR.Error` approach is you can still use the `RR` just like any other RR. The error might or might not be relevant to you.

Thanks @gucci-on-fleek !

Issue #213, opened 2026-02-05
Arvancloud provider


AzureLeMoon:
Hey there,

I have finished an initial libdns implementation of the provider "Arvancloud" with the api refrence available [here](https://www.arvancloud.ir/api/cdn/4.0).
i will open a pull request shortly

thanks

PR #214, opened 2026-02-05
New provider: Arvancloud

AzureLeMoon:
The associated pull request for #213 
A provider implementation for [Arvancloud](https://www.arvancloud.ir/api/cdn/4.0)


AzureLeMoon:
there is a caveat here, since arvancloud doesn't allow multiple IP addresses for a singe A record in the free plan, and doesn't allow the user to create CNAME records that point to domains not registered on the same account. the libdns test wouldn't pass normally so i had to make some changes like changing example.com to a domain i had and changing the name on the A record tests. however with those change the test passed completly

gucci-on-fleek:
> arvancloud doesn't allow multiple IP addresses for a singe A record in the free plan, and doesn't allow the user to create CNAME records that point to domains not registered on the same account. 

Well that's _really_ weird.

> the libdns test wouldn't pass normally so i had to make some changes like changing example.com to a domain i had and changing the name on the A record tests. however with those change the test passed completly

Normally, I'd say that if a downstream provider needs to modify the testing code to make it work, then that probably means that we should fix the testing code upstream/here, but this is such an unusual case, so since you were able to get things working yourself, I'm inclined to leave things as-is upstream.

---

Anyways, everything looks good to me after a cursory look through the code, so thanks for your contribution!

gucci-on-fleek (approved):

AzureLeMoon:
Hey thanks for the review 

> Well that's _really_ weird.

Yea it is, i managed to speak with them about this and they agreed to change the A record thing( it was treated as if the user was using the provider-side load-balance feature which we aren't since we just want all of the IP addresses for that record), and allow example.com as an exception for the CNAME records. 

I will push the code to the correct repositories as soon as i have access to them.

mholt (approved):
Thanks! I agree with @gucci-on-fleek -- LGTM. Make sure any caveats are noted in the documentation, but yeah I'll make repos for you to push to.

Issue #226, opened 2026-08-09
eDNS (edns.de) provider


chris299:
Hi,

I'd like to claim the provider **eDNS** ([edns.de](https://edns.de)), a German registrar and DNS host. The implementation is finished and in use: <https://github.com/chris299/libdns-ednsde>.

One thing worth flagging up front, because it shapes the package: eDNS exposes a dedicated **DNS-01 challenge API** rather than a general zone-management API. It can add and remove TXT challenge records, and nothing else — in particular there is **no endpoint that lists the records of a zone**.

Consequently the package implements only `RecordAppender` and `RecordDeleter`. `RecordGetter` cannot be answered honestly and `RecordSetter` cannot be implemented without reading first, so both are deliberately *absent* rather than present-and-failing: Go interfaces are structural, so a stub returning "not supported" would still satisfy `RecordGetter` and move the failure from compile time into the middle of a certificate request. `certmagic.DNSProvider` needs only the two implemented interfaces, so Caddy is fully served.

For the same reason there is one documented deviation from the `DeleteRecords` godoc: the contract allows an empty value to mean "delete everything matching the name", which would require reading the zone. That case returns an explicit error instead.

Testing: unit tests run against a stub server with responses recorded from the live API and need no credentials; integration tests run against the real API with `EDNS_TOKEN` and `EDNS_TEST_ZONE`, and clean up after themselves. Beyond that, a Caddy plugin wrapper has obtained real certificates from Let's Encrypt staging over DNS-01, for both a plain host and a wildcard.

I'll open a pull request for in-line review next. Happy to adjust anything.

Thanks!

chris299:
Both repositories are populated and tagged `v1.0.1`:

- [libdns/ednsde](https://github.com/libdns/ednsde) — the provider
- [caddy-dns/ednsde](https://github.com/caddy-dns/ednsde) — the Caddy module

Unit CI is green on both, and the integration run against a live zone is green. The review points from #227 are answered there, and `libdnstest` against a partial provider is tracked separately as #228.

Thanks @mholt and @gucci-on-fleek.

PR #227, opened 2026-08-09
eDNS (edns.de) provider — for in-line review (#226)

chris299:
Implements a provider for **eDNS** ([edns.de](https://edns.de)), a German registrar and DNS host. Claimed in #226.

Per the [wiki](https://github.com/libdns/libdns/wiki/Implementing-a-libdns-package), this PR is here so the code can be reviewed in-line and is **not meant to be merged**. The package sits in its own module under `ednsde/`, so it does not affect the build or tests of `libdns` itself (`go list ./...` and `go test ./...` at the root are unchanged).

The working repository is <https://github.com/chris299/libdns-ednsde>, with CI green on every push.

## The one thing I'd like your call on

eDNS exposes a **DNS-01 challenge API**, not a zone-management API: it can add and remove TXT challenge records and nothing else. There is no endpoint that lists the records of a zone.

So `GetRecords` cannot be answered honestly and `SetRecords` cannot be implemented without reading first. I chose to **omit both methods** rather than provide stubs that return an error. The reasoning: Go interfaces are structural, so a stub still satisfies `libdns.RecordGetter` — it only moves the failure from compile time into the middle of a certificate request. Omitting them makes a caller that needs reading fail to build, at the call site, with a clear message. `certmagic.DNSProvider` requires only `RecordAppender` and `RecordDeleter`, so Caddy is fully served.

I'm aware this differs from [`libdns/acmedns`](https://github.com/libdns/acmedns), which faces the same limitation and implements all four with erroring stubs, and that `libdnstest` requires all four interfaces so it cannot be applied here. I went the other way because the four interfaces are defined separately, which reads like an invitation to implement the subset a provider can actually support — but I have no attachment to it. **If you prefer consistency with `acmedns`, say the word and I'll switch; it's a small change.**

There is one related deviation from the `DeleteRecords` godoc, documented in the code and README: the contract allows an empty value to mean "delete everything matching this name", which would require reading the zone. That case returns an explicit error.

## Behaviour, established against the live API

The provider's shape follows from things the vendor documentation does not state, which I verified by probing a real zone:

- The API does **not** prefix `_acme-challenge` itself; the libdns record name becomes its `subdomain` parameter verbatim.
- An empty `subdomain` is rejected with `400`, so a record on the zone apex (libdns `"@"`) must omit the field entirely.
- Error responses carry `"data"` as an empty **array** while success responses carry an object, so the envelope is decoded as `json.RawMessage` first.
- TTL is fixed at 300s and cannot be influenced; input TTLs are ignored and the returned records report what is actually in the zone.
- Several values may share one name, which is what a SAN certificate for both a domain and its wildcard needs.

Measured propagation on a live zone: a record on a name that did not exist before becomes visible on all authoritative nameservers after ~23s; removals take ~307s, almost exactly the record TTL.

## Testing

- **Unit tests**: `net/http/httptest` stub returning responses recorded from the live API — every result code (1–5) and every documented error (400 malformed action / bad value / bad JSON / missing parameter, 401, 405), plus retry behaviour and context cancellation. No credentials needed, run in CI on every push.
- **Integration tests**: against the real API behind a build tag, with `EDNS_TOKEN` and `EDNS_TEST_ZONE` so anyone can plug in their own credentials. They verify propagation against the zone's authoritative nameservers and clean up after themselves, including on failure.
- **End to end**: a Caddy plugin ([chris299/caddy-dns-ednsde](https://github.com/chris299/caddy-dns-ednsde)) obtained real certificates from Let's Encrypt **staging** over DNS-01 — one for a plain host, one for a wildcard — and served requests through them.

## Checklist

- [x] Adheres to the libdns godoc, with the one deviation described above
- [x] Godoc on every exported field, method and function
- [x] `json:"snake_case,omitempty"` tags; non-serialisable fields marked `json:"-"`
- [x] `go mod tidy` clean; one dependency, `libdns` itself
- [x] Tested, and testable by others with their own credentials
- [x] Standard library only for testing
- [x] Pure Go, builds with `CGO_ENABLED=0`; no shelling out
- [x] Configuration through struct fields; the package reads no environment variables
- [x] Thread-safe, no data races (`go test -race` in CI); no explicit provisioning
- [x] MIT licence
- [x] README covering configuration, credentials, caveats and the propagation timings

Happy to adjust anything. Thanks for maintaining this project.


gucci-on-fleek:
(sorry for the delay)

> So `GetRecords` cannot be answered honestly and `SetRecords` cannot be implemented without reading first. I chose to **omit both methods** rather than provide stubs that return an error. The reasoning: Go interfaces are structural, so a stub still satisfies `libdns.RecordGetter` — it only moves the failure from compile time into the middle of a certificate request. Omitting them makes a caller that needs reading fail to build, at the call site, with a clear message. `certmagic.DNSProvider` requires only `RecordAppender` and `RecordDeleter`, so Caddy is fully served.
> 
> I'm aware this differs from [`libdns/acmedns`](https://github.com/libdns/acmedns), which faces the same limitation and implements all four with erroring stubs […] I went the other way because the four interfaces are defined separately, which reads like an invitation to implement the subset a provider can actually support — but I have no attachment to it. **If you prefer consistency with `acmedns`, say the word and I'll switch; it's a small change.**

Both options seem reasonable to me:

1. Implementing all 4 interfaces, and just returning an error for the unsupported ones seems fine, because _all_ `libdns` users need to implement proper error handling since there are dozens of reasons why modifying DNS records can fail (incorrect API keys and the network being down should apply to nearly everyone).

2. Only implementing the interfaces that you actually support also seems fine to me, because the interfaces are separate for a reason, and it seems kinda pointless to add an implementation that just returns a "not supported" error. This could potentially be incompatible with some API users, but as long as this works with whatever _you_ are planning on using (Certmagic presumably), then this should be fine.

> and that `libdnstest` requires all four interfaces so it cannot be applied here.

Hmm, that seems like an omission on our part. Feel free to open a separate issue and we can look into adding support for this.

> There is one related deviation from the DeleteRecords godoc, documented in the code and README: the contract allows an empty value to mean "delete everything matching this name", which would require reading the zone. That case returns an explicit error.

Again, as long as this works with whatever software you are planning on using, this seems fine.

> - TTL is fixed at 300s and cannot be influenced; input TTLs are ignored and the returned records report what is actually in the zone.
>
> […]
>
> Measured propagation on a live zone: a record on a name that did not exist before becomes visible on all authoritative nameservers after ~23s; removals take ~307s, almost exactly the record TTL.

Unfortunate, but this seems to be the case with most providers.

ednsde/client.go
```diff
@@ -0,0 +1,195 @@
...
+// DefaultEndpoint is the production eDNS challenge API.
+const DefaultEndpoint = "https://dns-challenge.edns.de"
+
+// The only two actions the API understands.
+const (
+	actionAdd    = "addChallengeRecord"
+	actionRemove = "removeChallengeRecord"
+)
+
+// Result codes reported in data.result_code on an HTTP 200 response.
```
gucci-on-fleek, line 22: Is this comment correct? Because the API returning 200 status code for `resultNotFound` seems a little odd to me (404 would make more sense), but I have definitely used APIs that work like this before.

ednsde/client.go
```diff
@@ -0,0 +1,195 @@
...
+
+// do performs one API call, retrying transient failures.
+func (p *Provider) do(ctx context.Context, req apiRequest) (*apiData, error) {
+	body, err := json.Marshal(req)
+	if err != nil {
+		return nil, fmt.Errorf("encoding request: %w", err)
+	}
+
+	var lastErr error
+	for attempt := 1; attempt <= maxAttempts; attempt++ {
```
gucci-on-fleek, line 135: Using a `range` is more idiomatic here:

```suggestion
	for attempt := range maxAttempts {
```

This is zero-indexed versus the current one-indexed, but I don't _think_ that that should cause any issues here.

ednsde/provider.go
```diff
@@ -0,0 +1,229 @@
...
+			return true, nil
+		}
+	case actionRemove:
+		switch data.ResultCode {
+		case resultRemoved, resultAlreadyRemoved:
+			return true, nil
+		case resultNotFound:
+			return false, nil
+		}
+	}
```
gucci-on-fleek, line 161: It seems kinda odd to me to _not_ return an error if you try to add a record that already exists or if you try to delete a missing record, but from an idempotent perspective I suppose it makes sense. (I don't think that you need to change anything here, just making sure that you're aware of any possible implications.)

ednsde/provider_test.go
```diff
@@ -0,0 +1,456 @@
...
+		respond     func(http.ResponseWriter, int)
+		wantCalls   int
+		wantSubdom  string
+		wantOmitted bool // "subdomain" key must be absent from the JSON body
+		wantErr     string
+		wantOut     int
+	}{
+		{
+			name:       "record added",
+			zone:       "winkler.tel.",
```
gucci-on-fleek, line 100: I'd recommend using `example.com` or similar here instead, since this code could stick around 10+ years from now, at which point you may not be involved any more.

(Also FYI, your/this domain appears to be expired/expiring very soon)

ednsde/README.md
```diff
@@ -0,0 +1,140 @@
...
+expires, an ACME client's propagation check cannot find the name's authoritative
+servers and will report "not ready", however promptly this package did its job.
+Check it, and keep it low:
+
+```sh
+dig +short SOA example.com | awk '{print "negative TTL:", $NF}'
+```
+
+At 86400 a first issuance can stall for a day. At 300 it costs at most five
+minutes, once.
```
gucci-on-fleek, line 117: Unless eDNS.de itself caches the records for the SOA minimum, I doubt that this is correct. Caching/recursive resolvers will indeed cache negative entries for up to this long, but I'm almost certain that Let's Encrypt queries the authoritative nameservers directly, therefore ignoring any TTLs/caches.

gucci-on-fleek (approved):
Looks good to me; thanks! I've pushed this to https://github.com/libdns/ednsde and given you admin access; I don't have access to https://github.com/caddy-dns/, so @mholt will need to create the repo for the Caddy plugin for you.

mholt:
Thanks for reviewing Gucci! And @chris299 I've added you to a caddy-dns repo for this provider.

ednsde/client.go
chris299: Confirmed against the live API: all five result codes arrive as HTTP 200, with `"status": 200` in the body as well — the code only ever lives in `data.result_code`. The API keeps HTTP status for transport, auth and validation problems (400, 401, 405) and reports domain outcomes in the body. A 404 never appears.

It turns out to be useful. `removeChallengeRecord` for a value that cannot exist is answered 200 / `result_code: 5` and changes nothing in the zone, so the end-to-end test uses exactly that call as a free credentials check — which turns a wrong or unassigned token into a one-second failure instead of a fifteen-minute wait that looks like slow propagation.

ednsde/client.go
chris299: Taken, thanks. The backoff becomes `time.Duration(attempt) * retryBackoff` and the guard `if attempt > 0`, so it stays three attempts with the same two delays and one fewer `-1` to read past.

libdns/ednsde@4f60cee

ednsde/provider.go
chris299: Deliberate — and a caller can still tell, which is what made it defensible to me. `result_code 5` maps to "not applied", so that record is left out of the slice `DeleteRecords` returns, and the godoc defines that slice as the records which were actually deleted. Comparing lengths gives you the answer.

Making it an error would break the case libdns explicitly cares about: an ACME client cleaning up after a crash, or a second `DeleteRecords` for a value a first one already removed.

On the add side eDNS is idempotent itself — `result_code 2` is "identical call made before, nothing changed" — and the record does exist afterwards, so reporting it as applied is the honest answer.

One asymmetry worth naming, since it is the same code path: eDNS only removes records it created through the challenge API. A TXT record put on the same name through the web interface comes back as `result_code 5`, i.e. as "not there", although it is. That is in the README and in the `DeleteRecords` godoc.

ednsde/provider_test.go
chris299: Taken — the unit tests use `example.com` throughout now. The only place a real zone appears is `EDNS_TEST_ZONE`, supplied by whoever runs the integration tests.

libdns/ednsde@4874099

And thanks for the heads-up on the domain. It renews automatically; it is registered somewhere other than it is hosted, which is probably what made the whois record look alarming.

ednsde/README.md
chris299: You are right, and the README was wrong twice — the second time only because I went looking after your comment.

On your point: rather than argue from memory I read CertMagic's check. `checkDNSPropagation` sets `checkAuthoritativeServers := len(m.Resolvers) == 0`, then `lookupNameservers` finds the zone's NS records and `checkAuthoritativeNss` queries each of those servers directly ([dnsutil.go#L231-L305](https://github.com/caddyserver/certmagic/blob/v0.25.3/dnsutil.go#L231-L305)). With a default Caddy config there is no recursive cache in the path at all, so a cached NXDOMAIN cannot be what I measured. The SOA minimum keeps a paragraph of its own now, scoped to the case where it is real: `resolvers` configured, or a client that checks recursively.

My first replacement explanation was that eDNS answers a challenge name from a generation living for the record's 300 s TTL, so a second value on a name waits it out. A sibling implementation of the same API had measured ~301 s for exactly that. I wrote it up, and then the integration run failed on a *fresh* name with two values — nothing visible on either nameserver within 120 s — which that model does not predict either. Three measurements later:

| | |
| --- | --- |
| two values on one fresh name | 6 s — no slower than one value |
| two consecutive fresh names, one value each | 36 s and 36 s |
| the exact failing sequence, replayed locally | 36 s then 6 s, no failure |
| the same integration run, re-dispatched | 24 s, and 23 s / 0 s for the two-value name |

So it is neither per-name nor zone-wide caching. Publishing an add takes somewhere between 6 and 36 seconds from here, took over 120 seconds once from a GitHub runner for a shape that takes 6 seconds locally, and nothing in the provider influences any of it. The only figure that reproduces is the removal: ~307 s, which is the 300 s record TTL, so a removed answer is left to expire rather than pushed.

The README now gives that spread and says the mechanism is unexplained, instead of a tidy table that was wrong twice. The recommendation survives on a plainer argument: give the propagation check ten minutes, because a tight timeout cannot make eDNS faster and only converts a slow publish into a failed certificate. The integration budget went from two minutes to six for the same reason — it exists to show the provider's calls take effect, not to hold eDNS to a latency it never promised.

Both dead ends are written down so the next person does not walk into them. Twice a measurement landed near 300 s and there happened to be a plausible 300 s quantity next to it — first the zone's SOA minimum, then the record TTL — and both times proximity got read as cause. Thanks for pulling the first thread.

libdns/ednsde@2e19029, libdns/ednsde@83c91f7

chris299:
Thanks for the review, and for the repo, @mholt.

@gucci-on-fleek: I'll keep the two-interface version. It is what I use through CertMagic, and the compile-time failure is the property I was after — but it is good to have both options weighed out loud rather than assumed. `libdnstest` against a partial provider is now #228.

Your inline points led to three changes, all of them in the new repo rather than here: the `range` loop, `example.com` in the tests, and the propagation section of the README. That last one was the valuable one — you were right that the SOA minimum could not be the cause, and going after it turned up that my replacement explanation was wrong too. The details are in that thread; the short version is that eDNS's publish latency varies from 6 seconds to over two minutes for the same shape, so the README now reports the measurements instead of a mechanism, and the recommended timeout stands on the plainer argument.

[libdns/ednsde](https://github.com/libdns/ednsde) and [caddy-dns/ednsde](https://github.com/caddy-dns/ednsde) are populated and tagged `v1.0.1`, unit CI green on both and the integration run green against a live zone, so #226 can be closed.

One thing worth flagging plainly since these are your organisations' repos, @mholt: both carry a `workflow_dispatch`-only job that talks to the live eDNS API — the provider's integration tests, and an end-to-end run that obtains a Let's Encrypt *staging* certificate through the module and serves a request through it. They write TXT records under `_acme-challenge*` into a real zone and remove them afterwards, including on failure; neither runs on push. I have set the eDNS access token as a repository secret in each. Say the word and I will take them out.

ednsde/README.md
chris299: Follow-up, since I left this at "the mechanism is unexplained" a few hours ago and that is no longer true.

I built a rig that publishes records and watches them from two machines at once — a workstation in Germany and a GitHub runner on Azure — both querying the zone's authoritative nameservers by IP, with the record names derived from a shared run id so the two logs merge on absolute timestamps. Four records, a minute apart:

| Record | `ns4` 149.13.77.2 | `ns3` 154.56.104.3 |
| --- | --- | --- |
| 1 | **17 s** | 66 s |
| 2 | 17 s | 29 s |
| 3 | 65 s | **26 s** |
| 4 | 27 s | 27 s |

The two observers resolved the same two addresses and agreed on all eight moments to within three seconds. So the vantage point is not the variable, and the CI runner I had implicated was innocent — that was one more coincidence read as a cause, the third in this thread.

What is left is simply that the zone's two nameservers pick a change up independently, and either one can be several times slower than the other, in either order, on consecutive records. A burst of twenty queries per server every three seconds across a publish never caught a half-updated answer and never saw two SOA serials behind one address, so each server flips atomically — two servers, not a fleet drifting apart. The same burst also killed a tidy-looking signal from the first run: the SOA serial had bumped within five seconds of every add there, but across an add and a delete in the burst run it did not move at all.

Practical upshot for anyone reading the package later, and it cuts in CertMagic's favour: `checkAuthoritativeNss` returns on the first authoritative server that carries the value, so it usually rides the faster one. It does give up the round on an NXDOMAIN from an earlier-listed server, so a lagging server still costs time — hence still a generous `propagation_timeout` — but it is not paying the slowest server the way a check demanding unanimity would. My own integration test does demand unanimity, which is why it was the thing that failed at two minutes while a real issuance would likely have gone through. That is now said out loud where the budget is set.

libdns/ednsde v1.0.2 and caddy-dns/ednsde v1.0.2 carry the corrected write-up; the rig and both raw logs are kept so the numbers can be checked rather than believed.

I have also asked eDNS what publish latency one may actually expect and what it depends on. That will take a while, and I will not hold anything up for it — but if the answer is interesting I will come back and say so.

